PR #17, opened 2021-07-06
Update Fourbytes and Asset urls to be env variables

androolloyd:
Update the fourbytes and asset endpoints to accept a ENV variable.


androolloyd:
if these get merged i can stop using my forked repo in the otterscan + erigon package.

src/components/MethodName.tsx
```diff
@@ -16,7 +16,8 @@ const MethodName: React.FC<MethodNameProps> = ({ data }) => {
 
     // Try to resolve 4bytes name
     const fourBytes = _name.slice(2);
-    const signatureURL = `http://localhost:3001/${fourBytes}`;
+    const fourBytesURL = process.env.FOUR_BYTES_URL ? process.env.FOUR_BYTES_URL : "http://localhost:3001"
```
wmitsuda, line 19: I think it is better to extract this constant and export it in params.ts. Also in Typescript, you can use the ?? operator, e.g., process.env.FOUR_BYTES_URL ?? "http://localhost:3001". See: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing

Lastly, I think this won't in a production build, React requires env variables to be prefix with REACT_APP, as explained here: https://create-react-app.dev/docs/adding-custom-environment-variables/

Could you make those changes, pls?

src/components/TokenLogo.tsx
```diff
@@ -13,9 +13,10 @@ const TokenLogo: React.FC<TokenLogoProps> = (props) => (
 );
 
 const InternalTokenLogo: React.FC<TokenLogoProps> = ({ address, name }) => {
+  const assetsURL = process.env.ASSETS_URL ? process.env.ASSETS_URL : "http://localhost:3002";
```
wmitsuda, line 16: Could make the same changes as above?

wmitsuda (changes requested):
I liked your idea, but can you pls make some changes? Also, please don't remove the package-lock.json, it is meant to be versioned, can you restore it in the PR (github doesn't let me comment on deletions).

wmitsuda:
Also, feel free to contribute with your experience here: https://github.com/wmitsuda/otterscan/discussions/20

I'd like to make the user experience less painful for people who just want to run it.

wmitsuda:
I'll close this PR, since now all assets are distributed under 1 unique docker image.

Thanks for the effort anyway, feel free to open another issue if you want to suggest more improvements.

PR #21, opened 2021-07-07
Use https instead of git URLs for submodules.

mariuspod:
This simplifies dockerized builds with git submodules as needed in this PR:
https://github.com/wmitsuda/erigon/pull/1

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "4bytes"]
 	path = 4bytes
-	url = git@github.com:ethereum-lists/4bytes.git
+	url = https://github.com/ethereum-lists/4bytes.git
 [submodule "trustwallet"]
 	path = trustwallet
-	url = git@github.com:trustwallet/assets.git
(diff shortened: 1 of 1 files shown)


androolloyd:
yes this is a must for my package to build as well,

wmitsuda:
Out of curiosity, why do you guys prefer https? Does it have something to do with certain network restrictions were you work?

androolloyd:
no its because there are no ssh keys to pull content in the build envs.

mariuspod:
That was the easiest solution to the problem where no SSH keys are available inside the docker image when building it. AFAIK git requires you to have SSH keys to clone from `git://` URLs and I thought that's a bit overkill to generate SSH keys into the docker image just to achieve a checkout. Maybe there's an option in git to be less strict on that however which I'm not aware of. Using `https://` based cloning doesn't require SSH keys but just requires host key verification which is fine.

wmitsuda:
It makes sense, thanks for the explanation!

PR #25, opened 2021-07-08
WIP: Dockerfile changes for performant docker build copying of large …

mariuspod:
…directories.

For some reason my docker (20.10.2) on ubuntu couldn't build your Dockerfile.
It bailed out with 100% CPU on the COPYing of the submodules dirs `4bytes` and `trustwallet` :cry: 
As a temporary workaround I'm zipping the files before the docker build and extracting them inside the container which works perfectly and is fast. If you have an idea why that's the case we can also get rid of the zippery.

I've already tested it with my erigon PR and pushed an additonal commit for using the built docker image from your repo:
https://github.com/wmitsuda/erigon/pull/1/commits/32ef882e369575651d755a06d2b2d6892339a100


wmitsuda:
That is interesting... maybe it is related to these directories (mainly 4bytes) having a lot of files. I'll search if there is some best practice to follow in this type of situation.

wmitsuda:
BTW, did you notice if this happens during the 4bytes copying or the trustwallet copying? 4bytes repo is smaller, but it has a lot of very small files. trustwallet is less files, but they are bigger.

I suspect it has something to do with specifically with 4bytes.

wmitsuda:
Just ran `wc -l` here, 4bytes has precisely 269868 small files in just one directory.

mariuspod:
Yes, I'm only experiencing this with the 4bytes dir, maybe I'm also too impatient but it shouldn't take more than 5 minutes if cp takes only 10 seconds :joy: 

Thanks to you already merged the https submodule PR I've adapted the Dockerfile to copy the 4bytes signatures and trustwallet assets directly from the builder image. I checked the amount of files on the host and inside the image 4bytes dir they matched.

The resulting image has 166MB and you can try to build it like this:
```
export VERSION_TAG=feature/docker-experiments
docker build -t otterscan:docker-experiments \
    --build-arg VERSION_TAG=$VERSION_TAG \
    --build-arg REACT_APP_ERIGON_URL=$REACT_APP_ERIGON_URL .
``` 

Later on we can tag the docker image with the `VERSION_TAG` e.g. `v2021.07.01-otterscan` but now for testing it's hardcoded.

mariuspod:
Okay, just added a stock github workflow to create docker images and push them to ghcr.io automatically:

https://github.com/mariuspod/otterscan/pkgs/container/otterscan

That's pretty handy for testing and it's self-contained and near the source code imo.
As mentioned before the `VERSION_TAG` is still hardcoded but that could be changed I guess to reflect the current branch/tag.

Do you plan adding it to dockerhub or is ghcr also an option ?

wmitsuda:
I was planning to add it to dockerhub, TBH I didn't know about ghcr, is it a popular registry nowadays?

I'm planning to cut a release after rebasing today's Erigon release and publish a new Otterscan release also, but now offering the docker image as a official way for consuming the app, no more local builds if you just want to run otterscan.

mariuspod:
Well it depends a bit if you want to integrate it directly with a CI/CD workflow into github then ghcr is really nice.
Also I it's super nice that you get also the README from your github repo which is auto-included into the package description of the image. I like it :-)

PR #308, opened 2022-07-15
feat(nginx): enable rpc/ws proxying

sambacha:
IDK if you would want to merge this, this may be better left in the documents, if yo think so I can open a PR adding this config in the docs. 

This nginx.conf enables nginx to act as a service proxy / load balancer for the backing eth client. It also tunes the settings to generally handle more load.

I can be specific about the settings if you would like further clarification

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -1,32 +1,77 @@
-user  nginx;
-worker_processes  auto;
+user nginx;
+pid /var/run/nginx.pid;
+worker_processes auto;
+worker_rlimit_nofile 65535;
+error_log /var/log/nginx/error.log notice;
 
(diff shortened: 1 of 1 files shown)


wmitsuda:
I have mixed feelings about this, tuning may be good; not sure about proxing, it seems more a personal architectural decision, and I don't see a gain in making the otterscan's nginx depend directly on where the erigon node is located inside the network, we can let this decision open to be made by whoever is consuming this docker image.

Also, the only reason there is a nginx in this docker image is because it needs a webserver to serve the SPA, the image is already too opinionated IMO.

BTW, why did you open another port for WS? erigon serves both regular http and WS from the same port.

PR #503, opened 2023-02-07
DecodedLogSignature adaptation tuple

jaxonly:
The page will error when the type is tuple, and displaying it in full will cause it to be very long. It is displayed directly as ```tuple[] param_name```


wmitsuda:
hi, thanks! do you have an example tx so I can test it here? (I could search for one, but if you have a ready one it would ease my life :)

wmitsuda:
thanks, merged!

PR #662, opened 2023-03-14
Add instructions for running on a Flask server

### LINKED ISSUE #555: highlight.js downloads highlighting for all programming languages
Otterscan has to make hundreds of requests to download all the `chunk-` highlight.js files, which contain the keywords for all the different programming languages. This takes a long time on slow connections. For example, we don't need highlighting for AppleScript or BASIC.

Can we limit this downloading to just Solidity, Vyper, and other Ethereum languages?

sealer3:
This shows an easy way to run a production build without Docker. For a full production environment you should have a production WSGI server running too, but this is sufficient for a modest number of users.

Closes #555

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,7 +34,7 @@
         "@zxing/library": "^0.21.3",
         "chart.js": "^4.4.8",
         "ethers": "^6.13.5",
-        "motion": "^12.5.0",
+        "motion": "^12.6.2",
         "prettier-plugin-organize-imports": "^4.1.0",
         "react": "^18.2.0",
         "react-blockies": "^1.4.1",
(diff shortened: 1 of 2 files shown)


wmitsuda:
going through old PRs... do you think this is still relevant? if so, I think we can add it to some section in the new docs page I'm thinking about (outside this github repo)).

sealer3:
I think it's important that we support a lightweight way for users to run an Otterscan production build without needing to run Docker or even nginx. This solution is still relevant, though we might instruct users to run it via gunicorn, a lightweight WSGI server, to improve performance.

wmitsuda:
let's put this one on hold then, so we have less work migrating all docs to the new repo.

I'm doing that slowly as a side task, trying to write all docs from scratch and rethinking the docs structure.

PR #686, opened 2023-03-18
Stop file watcher from watching 4bytes assets

### LINKED ISSUE #556: Limit file watcher directories
Running `npm run start` with all the submodules cloned, I am hitting the limit for file watchers because of all the 4bytes files:

```
Error: ENOSPC: System limit for number of file watchers reached, watch '/otterscan/4bytes/signatures/0da45188'
```

Is there a way to disable watching the assets directories? Or run in release mode maybe?

sealer3:
Fixes #556

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -13,9 +13,7 @@ export default defineConfig({
   ],
   server: {
     watch: {
-      ignored: [
-        "**/4bytes/**",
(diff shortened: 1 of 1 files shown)


wmitsuda:
thanks for the contribution and sorry for the delay!

PR #687, opened 2023-03-18
Add token equivalence map to token price oracle

### LINKED ISSUE #605: Wrapped Ether price in token transfers list
It would be nice to include the Wrapped Ether prices (equivalent to ETH's price) in the token transfers layout. Currently there is no price for Wrapped Ether:
image

sealer3:
Fixes #605 by mapping the Wrapped Ether address to Chainlink's "Ether" address

### DIFF
--- a/src/usePriceOracle.ts
+++ b/src/usePriceOracle.ts
@@ -20,6 +20,22 @@ const FEED_REGISTRY_MAINNET: ChecksummedAddress =
 // The USD "token" address for Chainlink feed registry's purposes
 const USD = "0x0000000000000000000000000000000000000348";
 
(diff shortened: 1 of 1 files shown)


sealer3:
@wmitsuda Can we revisit this? Not being able to see wrapped ETH prices in USD is a little frustrating.

wmitsuda:
let's do it, rebase plz.

sealer3:
Done. Here is an example:
image

(Any Uniswap trade that converts between ETH and WETH will have equivalent ETH/WETH amounts.)

wmitsuda (approved):
good!

PR #727, opened 2023-03-26
Add support for ERC-721 token transfer logs

### LINKED ISSUE #723: ERC-721 token transfer support
Currently we have support for showing ERC-20 token transfers inside transactions. We should include support for ERC-721 token transfers, which can be detected using the `Transfer(address indexed _from, address indexed _to, uint256 indexed _tokenId)` event as described in the [EIP-721 spec](https://eips.ethereum.org/EIPS/eip-721#specification).

sealer3:
Fixes #723.

The code I wrote needs a lot of polish (hence the draft), but it does show ERC-721 transfers. The only distinction between ERC-20 and ERC-721 is the token ID is indexed also in ERC-721 transfers, so we just modify the topic length check to account for this.

ERC-20 and ERC-721 share the same topic because (uint256 amount) is the same type as (uint256 tokenId).

### DIFF
--- a/src/execution/transaction/TokenTransferItem.tsx
+++ b/src/execution/transaction/TokenTransferItem.tsx
@@ -9,6 +9,7 @@ import { useBlockNumberContext } from "../../useBlockTagContext";
 import { useTokenMetadata } from "../../useErigonHooks";
 import { useTokenUSDOracle } from "../../usePriceOracle";
 import { AddressContext, TokenTransfer } from "../../types";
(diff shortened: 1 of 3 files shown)


wmitsuda:
I'd wait a little longer to work on this PR because I'm almost finishing the first prototype for token indexing, which would give you an API to correctly identify if the token is an ERC20 or ERC721, among other attributes, so you would not need to do this hacky check.

sealer3:
I agree, it is better to have the complete interface first. I will come back to this when it is complete.

PR #847, opened 2023-04-26
Update block gas target formatting

sealer3:
On the London screen, show the percentage of the gas usage relative to the gas target, like "18,450,975 (+23.01%)" and remove the gas target column.

Use two decimal places when showing the basefee.


wmitsuda:
will merge this one, although I've hid the london link from homepage, I think we'll probably remove this page in future.

PR #856, opened 2023-04-28
Show contract bytecode in Contract tab

### LINKED ISSUE #767: Show contract bytecode on request
We should have a way to see the deployed contract bytecode from the Contract tab. We can consider having the user ask for the bytecode (e.g. by clicking a "reveal bytecode" button) instead of fetching it every time.

Whether this is the currently deployed contract code or contract creation code (same as data in contract creation tx) is a question we should think about.

sealer3:
Fixes #767 

Note that this shows the deployed bytecode.

### DIFF
--- a/src/execution/address/Contracts.tsx
+++ b/src/execution/address/Contracts.tsx
@@ -5,8 +5,10 @@ import React, { useContext, useEffect, useState } from "react";
 import ContentFrame from "../../components/ContentFrame";
 import ExternalLink from "../../components/ExternalLink";
 import InfoRow from "../../components/InfoRow";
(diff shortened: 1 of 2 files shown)


sealer3:
@wmitsuda I'd like to revisit this. Example use case: I am trying to clone an ERC-1167 smart contract and need to see the bytecode. Zilliqa already merged it into their fork, so clearly there is other interest in the feature.

wmitsuda (approved):

wmitsuda:
ok, let's do it.

I'm still a little skeptical about the usefulness of showing raw bytecode data in the UI and having it load up to 24KB just by opening the tab though.

PR #888, opened 2023-05-08
Hide undefined activation eligibility epoch

### LINKED ISSUE #887: Eligible Epoch still shown for pending_initialized validators
Similar to #840. The eligible epoch is not determined yet if the balance of the validator is not 32 ETH yet. For example, some validators deposit a smaller amount first, like 1 ETH.

sealer3:
Fixes #887


wmitsuda:
hm, could you rebase and push this PR again? it got conflicted after I merged the previous one that changed the same file.

sealer3:
> hm, could you rebase and push this PR again? it got conflicted after I merged the previous one that changed the same file.

Done!

PR #901, opened 2023-05-12
Rename vite config and bump vite-imagetools to 5.0.3

sealer3:


sealer3:
Closed in favor of better solution here: #919

PR #1040, opened 2023-06-30
Transaction state diff tab

### LINKED ISSUE #1005: State changes tab for transactions
It would be quite helpful to have a tab of state changes for a transaction, similar to Etherscan. For example:
https://etherscan.io/tx/0xeb5e0dc7eeb4226518e53bc536c83ecd1d6f20b66735ea2adf04e7d50ba337f9#statechange
image

If I understand correctly, this would probably require a new API but not a new index, since its 'just' a transaction trace presented in a different way.

sealer3:
Provides a graphical display of the state difference produced by the transaction. A cleaner approach might be to extend the existing trace model and/or components to accommodate a state diff. Modifications appreciated.

Closes #1005 when complete

### DIFF
--- a/src/components/ElementDiff.tsx
+++ b/src/components/ElementDiff.tsx
@@ -1,0 +1,31 @@
+import React from "react";
+
+type ElementDiffProps = {
+  oldElem: React.ReactNode | null;
+  newElem: React.ReactNode | null;
+};
+
+const ElementDiff: React.FC<ElementDiffProps> = ({ oldElem, newElem }) => {
+  return (
+    <div className="flex flex-col rounded overflow-hidden">
+      {oldElem !== null && (
+        <div className="bg-opacity-10 bg-red-500 p-2">
+          <div className="flex items-start">
+            <span className="text-gray-500 mr-4 font-data select-none">-</span>
+            {oldElem}
+          </div>
+        </div>
+      )}
+      {newElem !== null && (
+        <div className="bg-opacity-20 bg-green-300 p-2">
+          <div className="flex items-start">
+            <span className="text-gray-500 mr-4 font-data select-none">+</span>
+            {newElem}
+          </div>
+        </div>
+      )}
+    </div>
+  );
+};
+
+export default ElementDiff;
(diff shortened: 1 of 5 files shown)


JamesHinshelwood:
Just tried this locally, looks great, thank you!
image

sealer3:
Cosmetic issues have been fixed, native token balances are shown with the fiat equivalent, and storage values use the uint256 decoder by default:
image

wmitsuda (changes requested):
ok, I like it, we can merge it.

I have some UI suggestions though:

wmitsuda:
could we add some padding inside the boxes to give it some breath? particularly in those nested boxes, it feels more evident because of the information density.

image

wmitsuda:
since it is a diff, what if we replace the <previous> -> <new> notation with something more like github diffs?

image

image

I'd suggest removing the arrow icon (and saving some vertical space in the process), fill the background for the previous value with light red, and the new value with light green. And prefix those divs with "-" and "+" icons, I think it should be a more intuitive notation.

sealer3:
New look:

image

For balances, I think incorporating the net difference would be useful. Do you have any suggestions for how that could look?

wmitsuda:
it's becoming even cooler :)

could we reduce the vertical padding of the red/green boxes by 1 or 2 pixels? I think now there is too much vertical space.

I like the balances suggestion, what if if experiment with a delta (need to find a delta icon) section between the red and green boxes? (or after the green, don't know what looks better).

also, not sure if it is better to use a neutral color like light yellow background to this delta box (so we condition the user to learn yellow == delta) and the value incorporates the "+" or "-" sign indicating a positive/negative delta, or the bg color should also use some red/green variation to reflect the delta direction as well.

sealer3:
The vertical padding is reduced now. I positioned the balance difference to the right of the before/after, like this:

image

I think it is intuitive enough without a delta symbol or even a box around it.

wmitsuda:
padding looks good now.

the USD values in green look very distracting for me, I'd like to use a more neutral color, does the gray one which we use by default for chainlink prices look bad?

also, for the balance delta at the right, I feel it misses some UI elements connecting it to the diff at the left. would it be too complicated to draw some connections between the before/after and the delta box? (still thinking about the best layout, though)

I'd add an explicit "+" sign to the USD value there as well to make it consistent with the raw delta.

sealer3:
Here is what it looks like with a netural preset and an explicit plus sign on the fiat difference values:

image

wmitsuda:
I think the UI could still be improved, but I'm out of ideas now, so let's merge it and gather feedback.

wmitsuda (approved):

PR #1173, opened 2023-08-16
Add interface for executing read-only contract functions

### LINKED ISSUE #1105: Interface for executing view function calls for contracts with source code
For contracts with source code, allow the view/pure functions of the contract ABI to be expanded, exposing input fields for executing them using `eth_call`.

For view/pure functions that have no arguments (such as getters), call the function automatically on expansion.

In the future we can cover transaction execution, which which use the same components. Also prepare for graphing the result of the view functions over a range of historical block numbers (see #1054).

sealer3:
Fixes #1105


wmitsuda (approved):
lgtm

Issue #1184, opened 2023-08-21
Migrate to ethers v6


wmitsuda:
Umbrella issue just to remind us that we are doing that :)

wmitsuda:
Assigned that to you @sealer3 .

Feel free to link any child PRs/discussions here, also to push smaller PRs that you think will make it easier to deal with given this is a big migration.

sealer3:
Discussion related to the JSON-RPC formatter: https://github.com/ethers-io/ethers.js/discussions/4324

sealer3:
I couldn't find a similar JSON-RPC formatter class in ethers v6, so one is included in the PR. I added full typing to it, so it should be easy to adapt to other ethers types and verify compatibility with them.

jxom:
Feel free to take a look into viem.sh too ;)

PR #1185, opened 2023-08-22
Use local commify function from ethers v5

sealer3:
Adds the `commify` function from ethers v5 in preparation for upgrading to ethers v6, which removed it. Other alternatives, such as the locale-specific `Intl.NumberFormat.format` function, work with plain JavaScript numbers and are subject to floating point number precision issues.

Part of #1184.


wmitsuda (approved):
lgtm

PR #1189, opened 2023-08-26
Migrate to ethers v6

### LINKED ISSUE #1184: Migrate to ethers v6

sealer3:
Closes #1184.

Needs to be done:
- [x] Add MIT attribution for the portions of the Formatter class from ethers v5

Testing still required:
- [x] ENS
- [x] Internal transactions
- [x] USD values
- [x] Trace
- [x] Beacon chain
- [x] Search
- [x] Navigation (block/address transactions)

Most of this ought to be testable via test cases in the future.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,7 +32,7 @@
         "@types/react-highlight": "^0.12.5",
         "@types/react-syntax-highlighter": "^15.5.7",
         "chart.js": "^4.3.3",
-        "ethers": "^5.7.2",
+        "ethers": "^6.7.1",
         "highlightjs-solidity": "^2.0.6",
         "react": "^18.2.0",
         "react-blockies": "^1.4.1",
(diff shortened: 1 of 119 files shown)


sealer3:
I've identified one regression: `useLatestBlockNumber` does not seem to be functioning anymore, even though there aren't any errors. An `eth_subscribe` call (when using a WebSocket RPC server) is followed shortly by `eth_unsubscribe`. Clearly the calls are coming from `useLatestBlockNumber`'s useEffect initialization and cleanup, but it's unclear why the listener is ending early.

sealer3:
Note: Some components will crash storybook until https://github.com/storybookjs/storybook/issues/22452 is addressed. But this shouldn't block us from merging.

wmitsuda (approved):
lgtm

PR #1211, opened 2023-09-05
Fix block transaction fee calculation

### LINKED ISSUE #1062: Bug: Incorrect total transaction cost in txs view for many type 2 transactions
The transaction cost reported inside a transactions list view is calculated incorrectly. The transaction view is not affected and shows a correct calculation.

It appears the cost is incorrectly calculated as `(basefee + maxPriorityFeePerGas) * gas consumed` when it should be `(basefee + min(maxFeePerGas - basefee, maxPriorityFeePerGas)) * gas consumed`.

Example tx: 0xace25621d138e91ee8a93f341b2db7804f0330d06d3f9c2dfe82333ee193221c

* Transactions list view (incorrect, 1.521123719857222 ETH):
    * The displayed value is exactly equal to `(45.72356707 gwei + 47 gwei) * 16404931 gas`.

image

* Transaction view (correct, 0.771031757 ETH):
  * The correct value is found as `(47 gwei) * 16404931 gas`.

image

sealer3:
Closes #1062

### DIFF
--- a/src/useErigonHooks.ts
+++ b/src/useErigonHooks.ts
@@ -111,6 +111,17 @@ const blockTransactionsFetcher: Fetcher<
         throw new Error("blockTransactionsFetcher: unknown tx hash");
       }
 
+      let effectiveGasPrice: bigint;
+      if (t.type === 2) {
+        const tip =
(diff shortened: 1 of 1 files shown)


wmitsuda (approved):
nice catch, thanks for finding it

PR #1219, opened 2023-09-06
Show PendingItem when loading contract sources

### LINKED ISSUE #1218: Display some loading indicator while loading source code from sourcify
I noticed something while testing #1130 , when switching between source files, or during the first load, we show no load indicator, which leads to no user feedback is the file is big.

Example contract where it happens on mainnet: 0xE592427A0AEce92De3Edee1F18E0157C05861564

sealer3:
The `h-screen` div class helps prevent a page height decrease when you change sources, which would send you up the page.

Closes #1218.

### DIFF
--- a/src/execution/address/ContractFromRepo.tsx
+++ b/src/execution/address/ContractFromRepo.tsx
@@ -2,7 +2,8 @@ import React from "react";
 import { SyntaxHighlighter, docco } from "../../highlight-init";
 import { MatchType, useContract } from "../../sourcify/useSourcify";
 import { useAppConfigContext } from "../../useAppConfig";
-import PendingItem from "./PendingItem";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faCircleNotch } from "@fortawesome/free-solid-svg-icons";
 
 type ContractFromRepoProps = {
   checksummedAddress: string;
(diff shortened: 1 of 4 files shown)


wmitsuda (commented):
Could we keep the border while showing the pending indicator?

Because there is some padding the pending box looks like "floating" inside an empty space below the tab.

(I tried to grab a screenshot, but it is too fast... but you should be able to notice it by manual testing)

sealer3:
Yes, I added a border to the pending box. I wish there were a way to preserve the source code div size when changing sources, but the current approach at least keeps your viewport position.

wmitsuda:
It looks better with the border, but now the pending indicator looks strange, like a giant gray box which does not take all vertical space, see the screenshot below :)

Having a preallocated vertical space looks good because it gives the feeling that something will be loaded inside it. but the gray thing doesn't look good.

what if we use some centered (horizontally and vertically) load indicator with some animated icon? (open to suggestions)

image

sealer3:
I agree. I like the idea of one that's centered in the larger space, and I'll find a better fitting loading indicator for it.

wmitsuda:
I liked the spinning ball. But as a last adjusment I think we should reduce the vertical space, I think it is now taking the size of screen, maybe 1/2 screen would look better? (feel free to test what looks nicer)

What annoyed me is that I had to scroll a lot in order to see the spinning ball, I think we shouldn't make the user scroll a lot in order to realize it is loading something.

So, we should find some midterm between have it occupy some enough space to realize there is something that will be loaded there, but also don't take so much space that the user has to scroll a lot to see the load indicator.

wmitsuda:
"too much blank space now" :)

image

wmitsuda (approved):
now its looks nice!

PR #1220, opened 2023-09-06
Add ENS domain support to Read Contract

### LINKED ISSUE #1193: Allow ENS on address fields on "read contract" page
if the field is of "address" type, and the chain supports ENS (which means only mainnet), we should allow ENS names as input.

The name should be resolved before the call is bundled and sent to the node.

sealer3:
As with strings, you need to surround ENS domains with quotes unless the entire parameter is an address.

Closes #1193

### DIFF
--- a/src/execution/address/contract/ReadFunction.tsx
+++ b/src/execution/address/contract/ReadFunction.tsx
@@ -1,6 +1,13 @@
 import { FC, memo, useContext, useState, FormEvent } from "react";
 import { SyntaxHighlighter, docco } from "../../../highlight-init";
-import { FunctionFragment, Result, Interface, type ParamType } from "ethers";
+import {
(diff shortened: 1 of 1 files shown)


wmitsuda (commented):
Are you able to load the page for your test contract on sepolia?

http://localhost:5173/address/0xb0e9daBA710617b2C398b3987716B2F1f36bB228/readContract

I'm getting some errors when opening that address page:

image

sealer3:
@wmitsuda ethers does not have support for the function elementary Solidity type yet. See https://github.com/ethers-io/ethers.js/issues/4344

wmitsuda:
I think I'll merge this one, but plz create another issue to track the function stuff + reference the etherjs one so we remember to test it in the future @sealer3

Issue #976, opened 2023-06-07
Evaluate date-fns usage


wmitsuda:
Suggested by external contributor: https://github.com/date-fns/date-fns

PR #1226, opened 2023-09-11
Add address balance graph

### LINKED ISSUE #1054: Account balance graph
For various reasons, it is often useful to track changes in an account's ETH balance over time. We can simply query balances at previous block numbers to populate a graph of the balance.

- [ ] Add tab to address page
- [ ] Create graph class. This should be generic enough to be reused for mapping token balances and even individual storage slots in the future using `eth_getStorageAt`.
- [ ] Populate the graph using queries. Consider using a technique called "progressive presentation" by making requests in an interlacing pattern and immediately updating the graph when responses are received. This way, the graph appears sparse at first, yet it loads quickly and covers the entire range; as more responses are received, the details of the graph take form.

sealer3:
Adds a tab to the address page showing the changes in native currency balance over the previous 90 days.

Closes #1054 

- [x] Make graph class general enough for supporting ERC-20 tokens and Read Contract integer return values in the future
- [ ] Add decimals customization option to read contract interface


sealer3:
This makes use of `date-fns` package for displaying dates in the graph, which #976 also suggests.

sealer3:
The UI might need some polish, but the idea deserves attention. Think of the expressive ability users would achieve from being able to graph over time any variable that can be returned from a smart contract call.

PR #1254, opened 2023-09-20
Add favorite addresses feature

### LINKED ISSUE #1006: Allow users to mark addresses as favorite
Most obvious use case: star your own addresses.

Where should we store that info? local storage seems obvious, although browse purge policies may mess with that, should test it.

sealer3:
Closes #1006


wmitsuda:
note: I'm pretty sure this functionality would benefit from the labelling/tagging infrastructure we are doing, although it may not be saved in the same shareable place, so I would focus on doing that first, then we revisit/rebase this UI on top of that.

PR #1270, opened 2023-09-23
Add contract tabs for matched proxy contracts

### LINKED ISSUE #1251: Add support for ERC1167 minimal proxies on "read contract"
Currently, the "read contract" tab is not proxy aware.

If erigon has ots2 API enabled, there is a new set of methods that allow you to know if an address is an ERC1167 minimal proxy and its logic contract.

The read contract should be proxy-aware by detecting if the address is a proxy, and if so, check if the logic contract is verified and use their ABI instead.

For now only ERC1167 proxies are indexed, more types are coming, so we should make the UI kind of generic.

The methods will probably change names in future, and there may be a more generic "is proxy?" and "get logic contract" methods, but for now let's start by implementing ERC1167 support.

The methods are: `ots2_getAddressAttributes` and `ots2_getERC1167Impl`.

Examples:

```
curl -v -H "Content-Type: application/json" --data '{"jsonrpc": "2.0", "method": "ots2_getAddressAttributes", "params": ["0xffc0e7288b285cfc0c849de17e2bdec9e6b1c41c"], "id":1}' <rpc-url> | jq -C
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "erc1167": true
  }
}
```

```
curl -v -H "Content-Type: application/json" --data '{"jsonrpc": "2.0", "method": "ots2_getERC1167Impl", "params": ["0xffc0e7288b285cfc0c849de17e2bdec9e6b1c41c"], "id":1}' <rpc-url> | jq -C
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": "0xd332254f274cc65aa11178b74734e2992b8f349e"
}
```

sealer3:
Closes #1251, and also displays the source code of the proxy's implementation address in a new tab.

Things to consider:
* Having separate tabs for a proxy Contract and proxy Read Contract tab is great for when other proxy types are supported, since proxy contracts might be verified and add functions on top of the implementation contract's functions.
* On the other hand, it takes up space in the header
* Maybe we should be clear about which type of proxy is matched and give a link to the implementation contract somewhere, e.g. in the proxy's Contracts tab.

### DIFF
--- a/src/execution/AddressMainPage.tsx
+++ b/src/execution/AddressMainPage.tsx
@@ -6,6 +6,7 @@ import {
   useParams,
   useSearchParams,
 } from "react-router-dom";
+import { getAddress } from "ethers";
 import { Tab } from "@headlessui/react";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faCircleNotch } from "@fortawesome/free-solid-svg-icons";
@@ -27,6 +28,10 @@ import { useAddressOrENS } from "../useResolvedAddresses";
 import { useSourcifyMetadata } from "../sourcify/useSourcify";
 import { ChecksummedAddress } from "../types";
 import { usePageTitle } from "../useTitle";
+import {
+  useAddressAttributes,
+  useERC1167Impl,
+} from "../ots2/usePrototypeTransferHooks";
 
 type AddressMainPageProps = {};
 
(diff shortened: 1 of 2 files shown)


wmitsuda:
do you have some contract address I could use to test it (on sepolia or mainnet)? otherwise I can search for it

sealer3:
@wmitsuda 0x9Ba02A9919beB4EA7B409B8528A675ce86dA5469 on mainnet is one example.

wmitsuda:
uh, it got a conflict because of another merge, could you rebase plz?

wmitsuda:
> 0x9Ba02A9919beB4EA7B409B8528A675ce86dA5469

that's strange, was able to get Sourcify working again, but this address seems to not be verified, see:

image

wmitsuda:
do you see something different in your env @sealer3 ?

sealer3:
@wmitsuda No, I believe I verified the implementation contract myself recently, so it may not be distributed via IPFS yet. Try using the HTTP Sourcify repo.

wmitsuda:
can't find it in the HTTP repo as well... maybe they restored the repo from a backup?

sealer3:
The implementation is at https://repo.sourcify.dev/contracts/full_match/1/0xE7fd730dD22d4C834CF6d5b8D539de8b14567AF4/metadata.json

and it should look like this in the UI:
image

wmitsuda:
now I see it, for some reason it is in the http repo, but not published in ipns. And because the ipfs chrome plugin intercepts published websites, when trying to access the http repo I was being redirected to the ipns one, I had to disable ipfs completely.

it looks good, it is working :)

but I think we can do some UI polishing here. I think I'll close the release 2.1 first with only withdrawals and postpone the merge of this one to the next release so we can think of some improvements.

I'll post my suggestion here.

sealer3:
Yes, and we should probably also tell users somewhere what type of proxy it is and the implementation address.

PR #1306, opened 2023-10-04
Replace 'network' key with 'name' in ChainInfo

### LINKED ISSUE #1253: Pull changes from op-otterscan in order to minimize the difference between forks
Ours frens from test-in-prod produced op-otterscan, which is the optimism-enabled version of ots suitable to work with op-erigon.

Their repo is: https://github.com/testinprod-io/otterscan/tree/op-erigon

This is an umbrella issue, we should try to pull their changes into Otterscan in order to minimize maintenance.

There may be changes that could be turned into parameters in Otterscan, others (like op-branded logo) could maybe be turned into a branding build time parameter, or a build time override (put a new logo in some specific place and it'll override the default one, for example).

Open to discuss all of it separately in smaller issues/PRs, also we don't need to do all of it at once, we can gradually evaluate smaller changes and link them here.

sealer3:
Removes the `network` key from ChainInfo, which doesn't seem to be used anymore by ethereum-lists. Adds `name` instead and applies it to the chain ID warning header if supplied.

Closes #1253

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -57,12 +57,14 @@ Some components in the user interface can be customized in the config under the 
 ```json
 {
   "branding": {
-    "siteName": "Otterscan"
+    "siteName": "Otterscan",
+    "hideAnnouncements": false
   }
 }
 ```
 
 * `siteName`: Sets the name displayed on the home page, header, and page titles.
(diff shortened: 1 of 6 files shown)


sealer3:
I also added the `hideAnnouncements` option to the config to complete the "common differences with op-otterscan" issue.

PR #1345, opened 2023-10-10
Update run-nginx.sh - exec in front

dan-aztec:
hope this will help with handling graceful termination from docker!  currently takes ~8s to shutdown on my `docker-compose down`, while other images shutdown instantly

image


wmitsuda:
chatgpt was right! I can reproduce it here, merging it!

Issue #1192, opened 2023-08-28
Better UI for inputting tuples on "read contract" view


wmitsuda:
not high priority, but we could present a nested set of inputs for individual tuple fields, so user can input each tuple element individually and we bundle and submit them together.

screenshot showing how it is currently:

<img width="693" alt="image" src="https://github.com/otterscan/otterscan/assets/28685/89f420a5-94aa-4f15-ac0d-1b8ed13b8e62">

wmitsuda:
that could be a good one to work one @sealer3 

feel free to break it into smaller PRs if you judge that's that case, like tuple, array, array of tuples, etc... :)

wmitsuda:
added tag to hightlight array params support on next changelog.

PR #1352, opened 2023-10-12
Split tuple inputs in read contract interface

sealer3:
Creates a new component that handles a list of parameters. This design makes recursive tuple input possible. Partially addresses #1192 since only base types of "tuple" are supported - arrays will require an interface for adding and removing elements.

### DIFF
--- a/src/execution/address/contract/FunctionParamsInput.stories.tsx
+++ b/src/execution/address/contract/FunctionParamsInput.stories.tsx
@@ -1,0 +1,56 @@
+import { Meta, StoryObj } from "@storybook/react";
+import { Interface } from "ethers";
+import FunctionParamsInput from "./FunctionParamsInput";
+
+const meta = {
+  component: FunctionParamsInput,
+} satisfies Meta<typeof FunctionParamsInput>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+const makeArgs = (fragment: string) => {
+  const intf = new Interface([fragment]);
+  return {
+    params: intf.fragments[0].inputs,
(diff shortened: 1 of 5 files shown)


wmitsuda:
I think this is a good opportunity to make some storybook stories for these components before we have too much of them to test.

Could you write some stories for the tuple case and also for the individual ones so we can easily test/inspect them in the storybook UI?

I think the component to be test could be the label + input box below it. Then we could easily see if tuples, arrays, etc, render the subtypes correctly.

sealer3:
I've added stories for this new component. Note that the tuple types will not be formatted correctly in the stories due to this bug in ethers: https://github.com/ethers-io/ethers.js/issues/4329

The JSON contract ABI returned from Sourcify is immune to this somehow.

wmitsuda:
oh, fuck... anyway, rebase again plz? I think it got conflicted after I merged the prettier PR.

sealer3:
Done

wmitsuda:
note to self: a good contract to test it on mainnet is: 0x1111111254EEB25477B68fb85Ed929f73A960582

wmitsuda:
I'd like to make one last UI adjustment given the recent improvement to color param names/types: see in the screenshot below, in this case, let's replace the "tuple....." with a similar colored string "tuple(...) order" (replacing the giant parameter list with an ellipsis because the params are now described individually).

image

sealer3:
How it looks now:

image

wmitsuda:
it looks good. there is still something missing linking the tuple to the inner params, maybe a different color or extra UI element, I'll think about that in another issue.

Issue #1348, opened 2023-10-11
Support new indexer: fee recipient


wmitsuda:
There is a new indexer for 2.2-ALPHA, given an address, returns a paginated view of all blocks it is the miner/block producer/fee recipient.

It works similar to the withdrawals search we've implemented in the past release, the method pair is: `ots2_getFeeRecipientCount/ots2_getFeeRecipientList` (name may change) and it is already available in our mainnet test server.

Le's implement a new tab in the address page, just after "Withdrawals".

Sample method call/returns:

```
curl -v -H "Content-Type: application/json" --data '{"jsonrpc": "2.0", "method": "ots2_getFeeRecipientCount", "params": ["0x690B9A9E9aa1C9dB991C7721a92d351Db4FaC990"], "id":1}' <rpc-url>
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": 385149
}
curl -v -H "Content-Type: application/json" --data '{"jsonrpc": "2.0", "method": "ots2_getFeeRecipientList", "params": ["0x690B9A9E9aa1C9dB991C7721a92d351Db4FaC990", 385130, 5], "id":1}' <rpc-url>
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "blocksSummary": {
      "0x1179c71": {
        "blockNumber": "0x1179c71",
        "timestamp": 1696995695
      },
      "0x1179c73": {
        "blockNumber": "0x1179c73",
        "timestamp": 1696995719
      },
      "0x1179c77": {
        "blockNumber": "0x1179c77",
        "timestamp": 1696995767
      },
      "0x1179c7f": {
        "blockNumber": "0x1179c7f",
        "timestamp": 1696995863
      },
      "0x1179c83": {
        "blockNumber": "0x1179c83",
        "timestamp": 1696995911
      }
    },
    "results": [
      {
        "blockNumber": "0x1179c71"
      },
      {
        "blockNumber": "0x1179c73"
      },
      {
        "blockNumber": "0x1179c77"
      },
      {
        "blockNumber": "0x1179c7f"
      },
      {
        "blockNumber": "0x1179c83"
      }
    ]
  }
}
```

For now, displaying the block number column is enough, in the meantime I may do some refactorings in the node side and think about which more data to add in the return json.

sealer3:
@wmitsuda Is there anything else to add to this page?

wmitsuda:
[...]

PR #1357, opened 2023-10-14
Add tab to address showing blocks produced

### LINKED ISSUE #270: Feature: Mined blocks
It would be helpful to show a list of all blocks mined by a given address. This might be exposed as a tab on the address page of an address that has mined at least one block. The feature would be especially useful when you are mining blocks on a private testnet.

sealer3:
Other block explorers populate this list with number of transactions, basefee, and total rewards.

Closes #270, #1348


wmitsuda (approved):
Looks good, will merge it though the functionality is not complete because I want to add more block-related columns before we cut the next release.

I'll follow up in another issue.

PR #1360, opened 2023-10-14
Remove manual insertion of ens plugin; it is added automatically

wmitsuda:
Setting the static network to mainnet is giving an error because apparently the ENS plugin is already loaded automatically.


wmitsuda:
not sure if it was some recent change in ethers, but anyway it fixes it

image

sealer3 (approved):
Looks good!

Issue #444, opened 2022-12-01
Handle "/" search shortcut key properly


wmitsuda:
Sometimes the event is triggered, but the character is still "typed" in the search box, like the event didn't consume the key.

wmitsuda:
dup of #1361

PR #1361, opened 2023-10-15
Fix '/' character keyboard shortcut handling

### LINKED ISSUE #1359: Typing "/" to access the search bar is also adding the "/" character to the search
There is a global "/" shortcut key to direct the focus to the search bar.

It used to work correctly, but I remember sometimes it was not only giving focus, but printing the "/" character in the search box implying I was probably not handling the keyboard event correctly.

But now I realized it is almost always printing the "/" character, so it is completely messed up :)

sealer3:
Closes #1359 and #444

Note that the '/' character is still entered into the input field if the input field already has focus, as expected.

### DIFF
--- a/src/search/search.ts
+++ b/src/search/search.ts
@@ -321,9 +321,15 @@ export const useGenericSearch = (): [
   };
 
   const searchRef = useRef<HTMLInputElement>(null);
-  useKeyboardShortcut(["/"], () => {
-    searchRef.current?.focus();
-  });
+  useKeyboardShortcut(
+    ["/"],
+    () => {
+      searchRef.current?.focus();
+    },
+    {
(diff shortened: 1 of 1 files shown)


wmitsuda:
it works perfectly, but did you understand why? Because I didn't, even reading the docs :)

sealer3:
It simply uses [`Event.preventDefault()`](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) under the hood to stop the keyboard input event: https://github.com/arthurtyukayev/use-keyboard-shortcut/blob/e4dc3e3ec96c32f8a4e9541ea74be74f455098e5/lib/utils.js#L1

wmitsuda (approved):

PR #1362, opened 2023-10-15
Don't show validator links on withdrawals page without beacon client

### LINKED ISSUE #1358: Don't show validators as links on withdrawals page when there is no beacon chain API configured
As title says :)

If it is not configured, we can still display the validators column, but don't linkify the IDs.

sealer3:
Closes #1358 

The validator tag not use a link when `beaconApi` is not defined in the config. I made the `faUser` symbol darker so that it doesn't look like a link when there is no link.


wmitsuda (changes requested):
What if instead of creating a new component, we add a "disabled" property to `ValidatorLink` so we avoid duplicating code?

And add a new variation to storybook: https://storybook.otterscan.io/?path=/story/consensus-components-validatorlink--genesis-validator

wmitsuda:
I also think for the "disabled" state we keep the text as the default color, but make the icon gray-ish so it doesn't stand much in relation to the text (don't want it to steal user's attention from the rest of the page, it should work as a "companion" to the validator info, i.e., the index).

We can iterate over the icon color, but I think the current one is stealing the attention because it is cyan.

src/consensus/components/ValidatorLink.stories.tsx
```diff
@@ -20,6 +20,13 @@ export const RecentMainnetValidatorIndex: Story = {
   },
 };
 
+export const DisabledValidatorLink: Story = {
+  args: {
+    validatorIndex: 560_000,
```
wmitsuda, line 25: instead of duplicating RecentMainnetValidatorIndex parameters, it is better to extend it and add the new properties like the SlashedValidator below.

src/consensus/components/ValidatorLink.tsx
```diff
@@ -31,13 +27,34 @@ const ValidatorLink: FC<ValidatorLinkProps> = ({ validatorIndex, slashed }) => {
...
+    </>
   );
+
+  if (disabled) {
+    return (
+      <span className={`flex items-baseline space-x-1 font-blocknum`}>
+        {labelElement}
+      </span>
+    );
+  } else {
```
wmitsuda, line 44: this else is not necessary

wmitsuda:
colors look good now

wmitsuda (approved):
lgtm

PR #1364, opened 2023-10-16
Add tx navigation by position in block

### LINKED ISSUE #345: Add navigation buttons for in-block navigation on tx page
Allow navigate to prev/next tx inside block on transaction details page.

Similar to nonce navigation.

### LINKED ISSUE #839: Navigation arrows for same-block transactions
Otterscan has the very nice feature of navigating to the previous or next transaction from the same sender. There are many times when you want to navigate between different transactions in the same block. For example, maybe you see a pattern of transactions together that you want to observe, like several token mints in a row.

I propose we also include navigation arrows in the Block / Position row, so that we can navigate transactions in sequential order of inclusion in the block.

sealer3:
Closes #345
Closes #839

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -11,6 +11,9 @@ import { ChainInfoContext, useChainInfoFromMetadataFile } from "./useChainInfo";
 
 const Block = lazy(() => import("./execution/Block"));
 const BlockTransactions = lazy(() => import("./execution/BlockTransactions"));
+const BlockTransactionByIndex = lazy(
+  () => import("./execution/block/BlockTransactionByIndex")
+);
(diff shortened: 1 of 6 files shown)


wmitsuda (changes requested):
I liked the result, it actually provides more meaning to that bar that indicates tx position inside the block.

having said that, there are a few gotchas that we need to address.

one that I just caught while examining the implementation is that it is producing multiple entries in the browse history probablyy due to the refreshes, see:

image

in that example, I opened the tx details and navigated back 2 txs.

wmitsuda (approved):
lgtm. I wonder if we can reduce the screen flashing effect when navigating, but we can address that separately.

PR #1377, opened 2023-10-18
Improve readability of read contract page

wmitsuda:
I got annoyed while testing some stuff using the read contract page, I decided to play with some modifications myself :)

What I did:

- Tried to diversify colors and sizes, on big contracts the page looks like a big dump of black text, so we need something to guide the user.
- Changed the formatting of params: param type is red, param name is blue, I tried to mimic the colors of source highlighter.
- Applied the code font since it is a function parameter.
- For this to work, I changed the code to not use the format function anymore, instead broke down the components and formatted them myself.
- I intentionally didn't applied that formatting to tuples and arrays because we are going to implement a proper UI for them, and when that happens I don't think we'll display that giant function declaration.
- I applied a different color to the numbers in the ordered list of functions. Light gray because we want to say: this info is associated with the function, but don't pay much attention to it, the function name is what the user should look at.
- That way we make users' eyes scan the page for black text, which are the function names, then the colored parameters guide through what needs to be filled.

Sample image from mainnet:

image

@sealer3 plz review and if it's ok, plz merge so you can rebase/continue the other PR from these changes.

### DIFF
--- a/src/execution/address/contract/ReadContract.tsx
+++ b/src/execution/address/contract/ReadContract.tsx
@@ -51,7 +51,7 @@ const ReadContract: React.FC<ContractsProps> = ({
               {viewFunctions.length === 0 &&
                 "This contract has no external view functions."}
               {viewFunctions.length > 0 && (
-                <ul className="list-inside list-decimal">
+                <ol className="marker:text-md list-inside list-decimal marker:text-gray-400">
                   {viewFunctions.map((fn, i) => (
                     <ReadFunction
                       func={FunctionFragment.from(fn)}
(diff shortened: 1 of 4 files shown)


sealer3 (approved):
I like the separation in color between the param type and name. The red color at first seemed out of place when I used a contract that had no param names, but it fits in with the Solidity styler.

wmitsuda:
I think we can make a synthetic param name when it is not present, like param_0 or <anonymous> with a different format to reduce this feeling of something missing

wmitsuda:
Added a synthetic param name for missing ones like we do in the method inputs, also applied a slightly different format to it.

image

PR #1407, opened 2023-10-23
Fix several Rules of Hooks violations

### LINKED ISSUE #1406: Errors on address pages: hook count changes
Navigating between address pages seems to throw errors like this:

```
Uncaught Error: Rendered more hooks than during the previous render.
```

I'm able to reproduce this by navigating to an address page and then trying to navigate to an ENS address.

sealer3:
Closes #1406


wmitsuda (approved):
lgtm

PR #1416, opened 2023-10-24
Use DNS-based IPNS by default in Sourcify links

### LINKED ISSUE #1415: Replace hardcoded ipns name for Sourcify
Instead of `/ipns/k51qzi5uqu5dll0ocge71eudqnrgnogmbr37gsgl12uubsinphjoknl6bbi41p`, let's use the more human readable, officially supported, DNS name: `/ipns/repo.sourcify.dev`.

The DNS solves to `/ipns/k51...`, so it should not affect the end result, but only gives us better confidence that the code/docs are pointing to something official from Sourcify.

sealer3:
Closes #1415

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -31,7 +31,7 @@ VITE_CONFIG_JSON='
     }
   },
   "sourcifySources": {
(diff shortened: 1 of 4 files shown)


wmitsuda (approved):
I tested it locally, it is correct.

PR #1417, opened 2023-10-25
Show contract attributes on the address Overview tab

### LINKED ISSUE #1385: UI improvements for the read contract proxy support
This is a follow up of #1251 with some UI polishing:

- Rename "Contract as Proxy" tab to "Logic Contract".
- If the contract is detected as a proxy, let's add two new entries to the "Overview" tab, just  below "Contract Creator":
  - "Proxy type": in this case it's be "ERC1167 minimal proxy". There will be other when we implement it in the node side.
  - "Logic contract": the logic contract address with all UI decorations as always.

sealer3:
Also renames "Contract as Proxy" to "Logic Contract".

Closes #1385

### DIFF
--- a/src/execution/AddressMainPage.tsx
+++ b/src/execution/AddressMainPage.tsx
@@ -2,7 +2,6 @@ import { faQuestionCircle } from "@fortawesome/free-regular-svg-icons";
 import { faCircleNotch } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { Tab } from "@headlessui/react";
-import { getAddress } from "ethers";
 import React, { useCallback, useContext } from "react";
 import {
   Route,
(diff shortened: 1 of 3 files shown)


sealer3:
Example:
image

wmitsuda (changes requested):
I still prefer "Logic contract" as the description instead of "Proxy logic address", and it matches the naming of the tab (which act more as a shortcut to the contract source code).

Can we also make a totally unrelated improvement here? I just realized that because the balance row is also lazy loaded, now it is very noticeable the it "jumps", can we apply some pending to the balance value itself and let the row always be showing so it doesn't jump?

sealer3:
Sure. I've changed the title to "Logic contract", though I wonder if people would have a hard time making the connection that it's part of the proxy.

And this is what the balance looks like while it's pending:
image

You might notice that the PendingItem component I used is slightly darker than the PendingResults below it.

wmitsuda (approved):

wmitsuda:
lgtm

PR #1422, opened 2023-10-27
Reduce horizontal margin on small screens

### LINKED ISSUE #1420: Decrease horizontal margin on mobile layout
Let's reduce the global horizontal margin by half on smaller screens in order to better use the scarce screen space.

See below a screenshot on a iPad.

A little unrelated, but I think we should also remove the entire ETH price/gas box on smaller screens because it is causing a layout break and I really don't know where to put it.

Related to #1252 

image

sealer3:
On medium width and smaller, cuts the horizontal margin by half.
On medium width and smaller, hides the price box.

There is still a layout break near the medium width cutoff due to the price box, but I'd rather not hide the price box at higher widths; at large width the price box and search are fully expanded.

We potentially could also change the price box so that it squishes at a slower rate than the search field.

Closes #1420.

### DIFF
--- a/src/Header.tsx
+++ b/src/Header.tsx
@@ -19,7 +19,7 @@ const Header: FC = () => {
   return (
     <>
       {isScanning && <CameraScanner turnOffScan={() => setScanning(false)} />}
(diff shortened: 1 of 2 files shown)


wmitsuda (approved):
it looks good, let's merge that so that I can test in a real device.

PR #1423, opened 2023-10-27
Revert "Bump @types/react from 18.2.31 to 18.2.32"

wmitsuda:
Reverts otterscan/otterscan#1418


wmitsuda:
The new types are causing compilation errors, it may be just the types are wrong.

PR #1429, opened 2023-10-28
Resolve verified contract names in decorated addresses

### LINKED ISSUE #1424: Use the contract name from Sourcify instead of the plain address when it is available
Like we use the ERC20 name() or the ENS name when this info is available, if the contract is verified on Sourcify and their main name is available from the metadata, we should display it instead of the address.

In the example below, we logic contract would be displayed as "WalletSimple" and the contract creator would be displayed as "WalletFactory".

That way we incentivize contracts to be verified instead of relying on manual labelling.

Regarding the label priority, I think it should be after the hardcoded labels we have (the 1.json files), i.e., if someone wants to override whatever name comes from Sourcify, he should be allowed by hardcoding the name in the json files.

Additionally, I'd like the text color to be displayed in the same "blue/purple-ish" color from Sourcify logo as a way to subtly signal in the UI that the name is coming from Sourcify metadata. (I think that'll look cool, but let's see it)

Feel free to refactor the current address component in order to support this new variation and keep the code maintainable.

image

sealer3:
Currently looks like this:
image

Closes #1424

Keep in mind that if Otterscan gets popular, this feature might be abused.

### DIFF
--- a/src/execution/address/renderer/VerifiedContractName.stories.tsx
+++ b/src/execution/address/renderer/VerifiedContractName.stories.tsx
@@ -1,0 +1,33 @@
+import { Meta, StoryObj } from "@storybook/react";
+import VerifiedContractName from "./VerifiedContractName";
+
+const meta = {
+  component: VerifiedContractName,
+} satisfies Meta<typeof VerifiedContractName>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
+  args: {
+    chainId: 1n,
+    address: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
+    linkable: false,
+    resolvedName: "WETH9",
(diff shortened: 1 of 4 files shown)


wmitsuda (changes requested):
cool, I think the colors worked really well.

can you make a storybook story for the new renderer? I made one for the ENSName component you can use as a template.

sealer3:
I made the storybook Story, but it won't run until https://github.com/storybookjs/storybook/issues/22452 is resolved.

wmitsuda:
> I made the storybook Story, but it won't run until [storybookjs/storybook#22452](https://github.com/storybookjs/storybook/issues/22452) is resolved.

oh, f*ck... I don't mind then, will merge it because this feature looks cool and must be released.

wmitsuda (approved):

PR #1439, opened 2023-10-31
Reduce margin starting at lg width

sealer3:
Reduces the margins on the sides starting at the lg width as suggested in #1420.


wmitsuda (approved):
I'll merge and test it before closing the parent issue.

PR #1443, opened 2023-11-01
Use ABI of verified contracts to decode function selectors

### LINKED ISSUE #1430: If a contract is verified by Sourcify, recover the method selector from the ABI
This is another improvement request involving verified contracts.

If the contract is verified, we should recover the method selector from the verified ABI and use its method signature instead of decoding if from the 4bytes dictionary (which can have collisions).

See for example the following screenshot: we should display `swapETHForExactTokens` instead of `0xfb3bdb41` in the method column.

Also, as a subtle UI hint, let's use the same Sourcify-blue text color instead of black in this case.

image

sealer3:
Closes #1430

### DIFF
--- a/src/use4Bytes.ts
+++ b/src/use4Bytes.ts
@@ -14,6 +14,7 @@ import { RuntimeContext } from "./useRuntime";
 export type FourBytesEntry = {
   name: string;
   signature: string | undefined;
+  fromVerifiedContract: boolean;
 };
 
 /**
@@ -70,6 +71,7 @@ const fourBytesFetcher =
       const entry: FourBytesEntry = {
         name: method,
         signature: sig,
+        fromVerifiedContract: false,
       };
       return entry;
     } catch (err) {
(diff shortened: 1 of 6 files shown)


wmitsuda (changes requested):
Hm, I think you didn't apply the Sourcify text color to the method text :)

src/use4Bytes.ts
```diff
@@ -85,25 +86,49 @@ const fourBytesFetcher =
  */
 export const use4Bytes = (
   rawFourBytes: string | null,
+  address?: string,
```
wmitsuda, line 89: could you add a brief explanation in the function doc above about the new address param? because it seems unintuitive why it exists and I'm pretty sure we'll struggle in the future to remember its meaning.

sealer3:
You're right. I'll add the Sourcify color and update the function doc.

wmitsuda (approved):
looks cool.

PR #1471, opened 2023-11-09
Add array entry interface and FunctionParamInput redesign

### LINKED ISSUE #1192: Better UI for inputting tuples on "read contract" view

sealer3:
Closes #1192 

Example:
image

Styled ParamDeclarations for arrays are not added yet. We can figure out how to represent those in a different issue.


wmitsuda (changes requested):
I'd like to request some minor UI adjustments to make it look nicer, see screenshot below:

- Baseline of "Remove" button text should be aligned other text's baselines.
- "Add Element" button left border should be vertically aligned with the parameters above.

Screenshot 2023-12-06 at 00 41 17

wmitsuda:
when the array is empty (because user removed all elements), we should give some more vertical space between the array signature and the "Add Element" button.

Let's try to give the same spacing as between the "Add Element" and the "Query" button in the screenshot.

And let's apply the same coloring/formatting to the `bytes32[] orderHashes` sig as we did for other types: `bytes32[]` (or `anytype[]`) in red, param name in blue.

if it is an array of tuple, I think we can show it as `tuple(...)[]`.

image

wmitsuda:
and as this is a complex component and I'm pretty sure we'll break it someday while refactoring it, I think we could include as testcase for it in the smoke tests, trying to interact with some mainnet contract and try to add elements, delete, etc.

I'm testing with the infamous `0x1111111254EEB25477B68fb85Ed929f73A960582` contract, but it is too slow to load, need to find some other we can use as test subject.

sealer3:
Here's what it looks like now:
image
image

I've also aligned the Query button for the other param types.

I haven't made the e2e tests yet, but I plan to create them for the devnet tests using a local HTTP server to serve the Sourcify assets. We can add some to the mainnet tests, too, but that introduces an additional availability requirement, namely that Sourcify's HTTP repo is online and accessible (we aren't rate limited).

wmitsuda:
> Here's what it looks like now: 

oh wait, for the baseline I meant the "bytes32..." text be aligned with the "Remove" button text (the "Remove" was a few pixels above previously, probably due to vertical centering or something) :)

sealer3:
Done:
image

wmitsuda (approved):
look cool!

wmitsuda:
I'm just wondering if e2e tests on devnet is worth it because of all the required setup.

sealer3:
There is some required setup, but it's very lightweight and will be consistent despite availability issues or changes to the mainnet node or Sourcify - in fact, they can help us figure out if these external services are down or have changed.

I think devnet tests will become more important when we wish to test how Otterscan reacts to changes in state or when we want to produce specific states that are difficult to come across and/or control on mainnet. For instance, testing a solution to #1595 on an address page would be straightforward on devnet (make 26 transactions to a fresh address), whereas on mainnet someone might send a 27th transaction to the address we choose and cause the test to fail, and on a permissionless network like Ethereum, anyone could do this to annoy us for a bit. On devnet, we get full control over the blockchain state. We can also test pending transactions.

But even better, external contributors who don't have access to (or don't want to run) a mainnet Erigon node can run devnet tests themselves, and the tests can run in GitHub Actions without needing access to secrets.

PR #1477, opened 2023-11-11
Use checksummed addresses in Token Balances tab

### LINKED ISSUE #1465: Addresses aren't checksummed on token balances tab
Token images, Sourcify labels, and more are not loading because the addresses aren't being checksummed when these requests are made.

sealer3:
Closes #1465


wmitsuda (approved):
ugh, thanks for catching this!

PR #1481, opened 2023-11-18
Add Cypress e2e tests and GA workflow

### LINKED ISSUE #1335: Initial infrastructure for end to end UI tests
Manual tests are starting to take a lot of time while modifying existing UI. So it's time to introduce some end-to-end testing in order to keep our mental sanity :)

- Evaluate some UI end-to-end testing infrastructure.
- Implement some smoke tests in order to evaluate it.
- Must integrate well with our github actions workflow.

sealer3:
Closes #1335 

In the future we might modify the workflow to cache the erigon binary so that it doesn't have to be re-compiled each time. Compiling does make it convenient to switch branches, such as to the ots2-alpha3 branch.

### DIFF
--- a/.github/workflows/end-to-end.yaml
+++ b/.github/workflows/end-to-end.yaml
@@ -43,5 +43,8 @@ jobs:
         uses: cypress-io/github-action@v6
         with:
           start: npm start
+          wait-on: "http://localhost:5173"
+          record: true
         env:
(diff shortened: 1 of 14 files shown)


wmitsuda (approved):
lgtm. will merge and move on, if there are adjustments will do in other PRs.

PR #1496, opened 2023-11-24
Show actual transfers in ERC20 Transfers tab

### LINKED ISSUE #1405: Show actual token transfers in token transfer tabs
The "ERC20 Transfers" and "ERC721 Transfers" tabs show the outer transactions during which token transfers happened involving the current address. The outer transactions often don't show which token is being transferred (unless it was transferred by directly interacting with the token contract), and they never show the transfer amount/NFT or whether the current address is the sender or recipient.

I think showing the token transfers themselves (token sender -> token recipient, token address, amount) would be more useful for tracking token transfers.

sealer3:
Closes #1405

We just fetch the token transfers with the `useTokenTransfers` hook that the transaction details page also uses.

Preview:
image

### DIFF
--- a/src/execution/address/AddressERC20Results.tsx
+++ b/src/execution/address/AddressERC20Results.tsx
@@ -6,6 +6,7 @@ import {
 } from "../../ots2/usePrototypeTransferHooks";
 import { usePageNumber } from "../../ots2/useUIHooks";
 import { PAGE_SIZE } from "../../params";
+import { findTokenTransfersInLogs } from "../../useErigonHooks";
 import { RuntimeContext } from "../../useRuntime";
 import { usePageTitle } from "../../useTitle";
 import { AddressAwareComponentProps } from "../types";
@@ -53,6 +54,7 @@ const AddressERC20Results: FC<AddressAwareComponentProps> = ({ address }) => {
           to: m.receipt.to,
           value: m.transaction.value,
           type: m.transaction.type,
+          tokenTransfers: findTokenTransfersInLogs(m.receipt.logs),
         }),
       ),
     [results],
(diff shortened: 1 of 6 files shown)


wmitsuda:
hm, I liked the idea, but I didn't like the rendering of the page jumping because each tx does a separate network call to get its transfers.

let me see if I can work in a proper API to support that. won't merge this one yet, but we can keep working on it.

sealer3:
I think an alternative to creating a custom API would be to delay rendering a page until the logs are retrieved. This would keep the page from jumping.

wmitsuda:
> I think an alternative to creating a custom API would be to delay rendering a page until the logs are retrieved. This would keep the page from jumping.

I don't know if it is possible without making the frontend code look ugly 😬

sealer3:
@wmitsuda I fixed the page jumping problem. I noticed the transaction receipt was being returned in the RPC response, and that is all we need. No extra overhead is required!

wmitsuda:
note to myself: I'm going to prioritize making ots2 work on e3, then we'll go back to this one.

wmitsuda:
hm, on a second thought I think we may merge this one.

I'm not completely satisfied with the UI yet because:

- Most of my ERC20 transactions are 1 single token transfer, which generates 1 extra row for each tx, it feels like a waste of screen space. I don't have a solution for that yet.
- For that reason, I don't want to expand this solution to the main (the first tab) search, and the block txs search as well; for ERC20 it may be ok since we are labelling it "experimental", but in the long term I wish to have an unified search. but I don't see that coming soon, maybe after e3 we can improve on it.
- On the other hand, it seems useful to see this info.

anyway, I think we can merge this one, but keep it only for the "ERC20" experimental tab. but let's not give so much publicity to it because we will change the UI in the future, so I don't want users to be so attached to it.

I think this one can merge cleanly, but could you rebase just in case @sealer3 ?

sealer3:
I agree - this view is specific to token transfers, and there is probably a better approach for the main transaction results pages. It's now rebased.

wmitsuda (approved):

PR #1504, opened 2023-11-27
Prevent SectionHighlighter from truncating by default

### LINKED ISSUE #722: Reconsider truncate class for certain decoded parameters
The `truncate` class, used in SelectionHighlighter, causes some parameters to get cut off in the decoded parameters table, the overflow replaced with an ellipsis. For instance, you cannot see an entire long `bytes` parameter because it is always cut off by this class.

There should be some way to see the entire decoded parameter. Showing the entire parameter also makes it easier to copy it to the clipboard. Maybe this is done by removing the `truncate` class from these decoded parameter rows.

### LINKED ISSUE #1500: Decoded bytes value should not be truncated/ellipsis
see mainnet tx: http://localhost:5173/tx/0x6d19a97f8ece4f746fbbb0b64ea4ba79500727ba95773cc3518de864184dd18a

the inputs[] should not be truncated, instead they should wrap and occupy all necessary lines so the user can see all bytes.

image

sealer3:
Closes #1500 
Closes #722 

Now:
image

Also, with this PR the "44 bytes" text shown in this picture no longer can be selected, which makes it easier to copy the bytes values without the extra length label.

### DIFF
--- a/src/components/AddressHighlighter.tsx
+++ b/src/components/AddressHighlighter.tsx
@@ -12,13 +12,15 @@ const AddressHighlighter: React.FC<AddressHighlighterProps> = ({
   address,
   children,
 }) => (
-  <SelectionHighlighter
-    myType="address"
-    myContent={address}
-    selector={addressSelector}
-  >
-    {children}
-  </SelectionHighlighter>
+  <div className="truncate">
+    <SelectionHighlighter
+      myType="address"
+      myContent={address}
+      selector={addressSelector}
+    >
+      {children}
+    </SelectionHighlighter>
+  </div>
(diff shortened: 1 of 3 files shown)


wmitsuda (changes requested):
be value truncation/overflow seems ok now.

but I'm seeing a flickering effect when hovering over any "highlightable" content, i.e., try to move the mouse over the from address on tx details page, you'll see the hightlighter box appearing/disappearing quickly, I didn't notice it before, so I believe it may be a side effect of this PR?

wmitsuda:
I'll merge it :)

PR #1558, opened 2023-12-01
Don't replace the @ symbol in Sourcify sources

### LINKED ISSUE #1295: Sourcify sources with @ symbol do not load
This happens when the source comes directly from Sourcify servers; in Otterscan, the path with an @ symbol is replaced with an underscore, so the contract source doesn't load. Example contract on mainnet: 0xb27308f9F90D607463bb33eA1BeBb41C27CE5AB6

There are other symbols that don't get replaced anymore, and if a replacement is needed, it is written to `path-translation.json` as per https://github.com/ethereum/sourcify/pull/1132

sealer3:
Closes #1295

### DIFF
--- a/src/sourcify/useSourcify.ts
+++ b/src/sourcify/useSourcify.ts
@@ -245,7 +245,7 @@ export const useContract = (
   type: MatchType,
 ) => {
   const sourcifySources = useSourcifySources();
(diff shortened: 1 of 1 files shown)


wmitsuda (approved):
lgtm

PR #1561, opened 2023-12-02
Fix transaction page navigation

### LINKED ISSUE #1560: Transaction page: navigating to Trace tab and back is broken
The tabs somehow are trying to navigate relative to the current URL and not the transaction URL, so if you click the Trace tab enough times, you get

`/tx/0xaf226ef77eae07429d123f3779ec747580dcf1b621014acf9085fcb426a783b0/trace/trace/trace/trace/trace/trace/trace/trace/trace/trace`

The cause seems to be from this update to react-router-dom: https://github.com/otterscan/otterscan/commit/c96c36a52a76f09cafacd3a73b671fd9785a2dbd

The reason was that we relied on so-called "buggy behavior" of the router, and so we need to update to react-router-dom v6.20.1 and/or figure out the correct way of referencing things relative to the parent page, not the current page: https://github.com/remix-run/react-router/issues/11052#issuecomment-1828470203

- [x] ~Fix navigation~ Update to v6.20.1
- [x] Add end-to-end test

According to the React Router issue I linked to above, the authors of React Router will release a statement with explanations about how to migrate to the new behavior.

sealer3:
Closes #1560


wmitsuda (approved):
thanks for catching this one!

wmitsuda (changes requested):
actually, it conflicted because I already bumped the package on devel, can you rebase and solve the conflicts, please?

sealer3:
Done!

wmitsuda (approved):

PR #1580, opened 2023-12-07
Show user notices and dev docs in decoded functions

### LINKED ISSUE #1472: Show user/dev doc for entire functions
Currently we show user/dev docs from verified contracts in the parameter names through a ? button that the user can click on. We should also have this for entire verified functions, which often have notice/dev/return tags attached to explain what the function is doing. Sometimes these describe what the inputs mean and are necessary to learn what the function is doing.

sealer3:
Closes #1472

Preview (user + dev docs):
image

I didn't add much styling other than the dev label. The spacing might need adjustment.


wmitsuda:
I was just wondering if it is better to show that info in the input section or show it together with the "transaction action" section above.

what do you think? in the following example we would show the text: "Swaps `amountIn` of one token for as much as possible of another token" as an explanation of the `exactInputSingle` method.

image

wmitsuda:
for the record, I'm using the mainnet tx `0xa6d0a9ea478a7dc68db5749306f81fbc380825b53d088b878df83aef835838e7`   as a test.

sealer3:
In my opinion, we should add the complete function signature near the top of the table and keep the documentation somewhere there, since whenever I read the input function parameters I also want to read the signature. Either way, I think you're right, the docs should go near the signature/function name. But in some cases you need the parameters to understand the docs, or want to reference the docs when reading the parameters.

wmitsuda:
The issue of adding the method signature is creating a readable string/UI for complex signatures (tuples, arrays, etc.).

On the other hand, what if we move the entire "Input data" section up, just after the "Transaction Action" section? and then move the function user/dev docs (not the params) to the "Transaction Action" section?

does it make more sense that way?

sealer3:
Would something like this work better? I really think there should be a function signature near the parameters so you can quickly see the function without having to scroll through the input parameters to see the types. We can use compact types for tuples, like Etherscan does:

image

Expanded to show the user doc:

image

Without compact types, the function signature looks more crowded but would be more convenient for developers:
image We could possibly show the full version when the help ? button is clicked.

wmitsuda:
I'm still not satisfied with the function sig proposal (idea makes senses, but the UI not good).

what if we merge the original proposal (user+dev docs) which we seemed to have reached consensus and discuss the function sig rework in another issue? (still no clear idea of good UI for that)

still, I think (user+dev docs for function) should be (for now) tied to the "transaction section" instead of the input block.

sealer3:
Sure. I'm not sure if we should put the text in a container or not, but here is how it looks like without one:

User:
image

Dev:
image

User+dev:
image

wmitsuda:
I was thinking about hiding the explanation by default and have a "?" icon to display it like what happens in the param section. And have the "?" icon be displayed only when the natspec exists.

but it looks way better now.

wdyt?

sealer3:
Okay, here's what it looks like now:

image

Open:

image

src/components/HelpButton.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { Switch } from "@headlessui/react";
+
+type HelpButtonProps = {
+  checked: boolean;
+  onChange: (newValue: boolean) => void;
+};
+
+const HelpButton: React.FC<HelpButtonProps> = ({ checked, onChange }) => {
+  return (
```
wmitsuda, line 14: I think we can simplify this code by removing the ` { return` block since we are returning only JSX

src/execution/transaction/decoder/DecodedParamsTable.tsx
```diff
@@ -16,41 +16,44 @@ const DecodedParamsTable: FC<DecodedParamsTableProps> = ({
...
-        <th className="col-span-8 pr-1">value</th>
-      </tr>
-      {!hasParamNames && (
-        <tr className="grid grid-cols-12 gap-x-2 bg-amber-100 py-2 text-left text-red-700">
-          <th className="col-span-12 px-1">
-            {paramTypes.length > 0 && paramTypes[0].name !== ""
-              ? "Parameter names are estimated."
-              : "Parameter names are not available."}
+}) => {
+  return (
```
wmitsuda, line 23: same here

wmitsuda (changes requested):

wmitsuda:
> Okay, here's what it looks like now:
> 
> image
> 
> Open:
> 
> image

the `?`icon look a little odd, I think that's because it is aligned with the text baseline instead of vertically centered (I can see that by inspecting the gridlines with chrome devtools).

wmitsuda:
but it looks way better overall!

wmitsuda:
also, need to find a solution to make it respect linebreaks, see i.e.: tx 0xc54acb4d67da6071290f06c4aaaa5b895a7341528b2171690d20a3c11feeea25 on mainnet

image

image

wmitsuda:
I'm wondering if we can assume the docs are always markdown and use some markdown renderer...

sealer3:
> also, need to find a solution to make it respect linebreaks, see i.e.: tx 0xc54acb4d67da6071290f06c4aaaa5b895a7341528b2171690d20a3c11feeea25 on mainnet

Unfortunately, the metadata file for that contract doesn't contain line breaks in the dev docs. Here's the metadata.json for that contract: 
https://repo.sourcify.dev/contracts/partial_match/1/0x8B227d72570D3EAd66014bCA8305CbeF7f90D1eE/metadata.json

I think the reason Solidity does this is that often lines are broken just to keep within a certain line width, but really we might want to open an issue in the Solidity compiler because two line breaks in a row clearly means there should be one there.

sealer3:
> I'm wondering if we can assume the docs are always markdown and use some markdown renderer...

The official NatSpec page has an example using backticks for variables too. Although the NatSpec page doesn't mention anything about Markdown, I think Doxygen, from which NatSpec was inspired, supports Markdown formatting, so it probably came from there.

It would be interesting to learn the extent to which Markdown is used in these NatSpec comments in smart contracts.

wmitsuda:
> > also, need to find a solution to make it respect linebreaks, see i.e.: tx 0xc54acb4d67da6071290f06c4aaaa5b895a7341528b2171690d20a3c11feeea25 on mainnet
> 
> Unfortunately, the metadata file for that contract doesn't contain line breaks in the dev docs. Here's the metadata.json for that contract: https://repo.sourcify.dev/contracts/partial_match/1/0x8B227d72570D3EAd66014bCA8305CbeF7f90D1eE/metadata.json
> 
> I think the reason Solidity does this is that often lines are broken just to keep within a certain line width, but really we might want to open an issue in the Solidity compiler because two line breaks in a row clearly means there should be one there.

ugh... or maybe they didn't know how to encode the linebreaks inside the json... that's unfortunate...

wmitsuda:
what if we assume ". " (dot + whitespace) is a paragraph? and optionally allow the user to view the raw content? (downside is making the component more complex)

sealer3:
> the `?`icon look a little odd, I think that's because it is aligned with the text baseline instead of vertically centered (I can see that by inspecting the gridlines with chrome devtools).

It might appear this way because the function name is size `text-xs` while the `?` icon is `text-sm`, but it's not explicitly aligned with the baseline (maybe through FontAwesome?). If I make the button `text-xs`, it will look like this:
image

sealer3:
> what if we assume ". " (dot + whitespace) is a paragraph? and optionally allow the user to view the raw content? (downside is making the component more complex)

I'm not sure - there are many NatSpec comments that have more than one sentence but don't create a new paragraph, such as a Uniswap execute tx on mainnet: 0xefcf7fcc800ec038c06cc8abfbb2240600aa44cd50031fb1575ecf87b5db691e

wmitsuda (approved):

wmitsuda:
ok, let's merge it. I think we can still improve UI overall (for both method and input params formatting), but let's think about it separatelly (once I get some better inspiration :) )

PR #1594, opened 2023-12-09
Display uint256 values with unselectable commas

sealer3:
This adds a component called DisplayInteger that renders an integer with comma separators which are not selected when the rest of the number is clicked. The number still is considered a complete word by the browser, so double-clicking a DisplayInteger that renders as "123,456,789" will select the string "123456789". Users therefore won't have to remove commas manually or do any extra work to select numbers.

This component could potentially be used in other areas of Otterscan where it's nice to show commas but where users might want to select text without the commas, such as in showing the current block number or validator.

I also added an attribute that enables a hint to the comma that marks the 18th decimal place in the form of a drop shadow:
image

I'm not sure if the shadow is really necessary. The idea was to add a non-obtrusive hint so you don't have to click the button to convert it to the "18 dec" view. This helps when there are many uint256 numbers being shown, and you don't want to click to convert all of them.

Thoughts?

### DIFF
--- a/src/components/DisplayInteger.stories.tsx
+++ b/src/components/DisplayInteger.stories.tsx
@@ -1,0 +1,34 @@
+import { Meta, StoryObj } from "@storybook/react";
+import DisplayInteger from "./DisplayInteger";
+
+const meta = {
+  component: DisplayInteger,
+} satisfies Meta<typeof DisplayInteger>;
+
+export default meta;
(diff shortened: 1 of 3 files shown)


wmitsuda (changes requested):
we can give it a try, but I found the "shadowy comma" 😂 a little noisy, let's remove the shadow pls.

wmitsuda (approved):

PR #1596, opened 2023-12-09
Add copy button to token addresses on Token Balances page

### LINKED ISSUE #1474: Add copy button to token addresses in Token Balances list
Similar to DecodedParamsTable, it should be easy to copy addresses of tokens in the Token Balances page.

sealer3:
Closes #1474

Example:
image

I wonder if this would be useful in other places token addresses appear, somehow without overcrowding the UI.

The main reason I wanted this feature was to copy token addresses into Read Contract input fields.

### DIFF
--- a/src/execution/address/TokenBalance.tsx
+++ b/src/execution/address/TokenBalance.tsx
@@ -1,5 +1,4 @@
 import { FC, useContext } from "react";
-import Copy from "../../components/Copy";
 import USDAmount from "../../components/USDAmount";
 import { useTokenBalance } from "../../ots2/usePrototypeTransferHooks";
 import FormattedBalanceHighlighter from "../../selection/FormattedBalanceHighlighter";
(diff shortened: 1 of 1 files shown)


wmitsuda (changes requested):
could you take a look at the `TransactionAddressWithCopy` component?

I really don't remember why I didn't use it inside `TokenBalance`, if I just forgot about it, or `TransactionAddressWithCopy` is doing more than we need in that page (hence needs more refactoring).

if you think `TransactionAddress` (which is wrapped into `TransactionAddressWithCopy` is doing too much and deserves a refactoring by itself, we can just keep your PR for now.

sealer3:
TransactionAddress does do a lot, but I think it's better to use TransactionAddressWithCopy than to duplicate code. Thanks for pointing me to that.

wmitsuda (approved):

PR #1605, opened 2023-12-15
Fix duplicate github actions execution

wmitsuda:


wmitsuda:
@sealer3 FYI: I think adding `feature/**` was my mistake as PRs are already executed by the pull_request/synchronize trigger, so removing it because the jobs were being executed twice each.

sealer3:
Are you sure? I don't see any examples of duplicated runs.

PR #1647, opened 2023-12-30
Use checksummed addresses in contract lists

sealer3:
Required for the Sourcify integration and other features to work with the addresses in contract lists


wmitsuda (approved):
thanks for catching that!

PR #1660, opened 2024-01-03
Add toggle for non-view functions in Read Contract

### LINKED ISSUE #1579: Option to show non-view functions that return values in Read Contract interface
Some Uniswap functions modify blockchain state but actually are meant to be used in contract calls, such as those in the V3 Quoter (0xb27308f9F90D607463bb33eA1BeBb41C27CE5AB6) and V3 Price Oracle (0x885a58B151Cfa0A1Af467bd246022DE6EBcd2626).

We should add a toggle to the Read Contract interface that adds the functions which modify blockchain state but return a value.

sealer3:
Closes #1579. Adds a toggle for non-view functions that have return values below the view functions:

Off (default):
image

On:
image

### DIFF
--- a/src/components/LabeledSwitch.stories.tsx
+++ b/src/components/LabeledSwitch.stories.tsx
@@ -1,0 +1,23 @@
+import { Meta, StoryObj } from "@storybook/react";
+import LabeledSwitch from "./LabeledSwitch";
+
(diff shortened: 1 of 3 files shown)


wmitsuda (approved):
lgtm!

PR #1681, opened 2024-01-12
Show function signatures

### LINKED ISSUE #700: Function signature not shown in decoded input data view
The function signature should be shown near the parameters in the decoded view of input data:

image

Even if we don't know the parameter names, we should always show the signature that was decoded right next to the decoded parameters.

### LINKED ISSUE #1674: Adding a function signature near the decoded parameters table
I strongly feel there should be a function signature near the parameters in the decoded parameters table in both transactions and traces.

Motivation:
* The function signature acts as a quick summary of the transaction action and gives extra context without even reading any input parameters
* Currently, the function name is separate from the parameters table, so if you wanted to see the name of the function, you have to find it above/outside the table
* There may be a handful of parameters, or parameters that are very large, making it an exercise to scroll and read all the parameter names and types from the table to figure out which function is called or which parameters are required to repeat the operation
* Easy access to the function signature so developers can test the call in some other context
* The function signature occurs at the bottom of the transaction page, so the user has to scroll already to see it, so adding the function signature doesn't block other content

Possible example we can talk about from #1850:
image

This version uses compact types like Etherscan does ([example](https://etherscan.io/tx/0x1c84c75cabeb054b7f57cc2af6d9ca9e5959ec9405f277271fbe78f0b4c7ae4d)), which limits the usefulness of the function signature but keeps the size limited. 

Showing full types like Blockscout does ([example](https://gnosis.blockscout.com/tx/0x4a2585f69af656eaf8b290404e9234c48629cd2261f710b3e18034a66e0a87e0)) is more useful, but might take up more space if the function is complex:
image

sealer3:
This PR adds a FunctionSignature component that shows a colored function signature and modifies the ParamDeclaration component to support full signatures in addition to short ones with abbreviated tuples.

Here is how it looks. I've added the help button here so that it appears in traces but of course we can remove that part if desired.
image

Something that might improve the look is to switch from `font-code` to `font-mono` for the parameters. I kept it at `font-code` since that is what `ParamDeclaration` uses, but I think `font-mono` would look better:
image

This probably varies per machine but at least for my default font it looks better :)

This is what it looks like (with font-mono) using 4bytes instead of a verified contract:
image

My intention was to give the decoded function signature a style similar to the decoded log signature which is present in the Logs tab:
image

Closes #1674
Closes #700
Closes #721

### DIFF
--- a/src/execution/address/contract/ReadFunction.tsx
+++ b/src/execution/address/contract/ReadFunction.tsx
@@ -221,7 +221,7 @@ const ReadFunction: FC<ReadFunctionProps> = ({ address, func }) => {
             func.inputs.map((param, index) => (
               <li className="mt-2" key={index}>
                 <div className="text-sm font-medium text-gray-600">
-                  <ParamDeclaration input={param} index={index} />
+                  <ParamDeclaration input={param} index={index} short={true} />
                 </div>
                 <FunctionParamInput
                   param={param}
--- a/src/execution/components/ParamDeclaration.tsx
+++ b/src/execution/components/ParamDeclaration.tsx
@@ -1,27 +1,57 @@
 import { ParamType } from "ethers";
-import { FC } from "react";
+import { FC, type ReactNode } from "react";
 
 type ParamDeclarationProps = {
   input: ParamType;
   index: number;
+  short?: boolean;
 };
 
-function getShortenedParamType(paramType: ParamType): string {
+function insertCommaEveryOther(components: ReactNode[]): ReactNode[] {
+  const result = [];
+  for (let i = 0; i < components.length; i++) {
+    if (i > 0) {
+      result.push(", ");
+    }
+    result.push(components[i]);
+  }
+  return result;
+}
+
+function getParamType(
+  paramType: ParamType,
+  short: boolean = false,
+): ReactNode[] {
   if (paramType.isArray()) {
     const end =
       paramType.arrayLength === -1 ? "[]" : "[" + paramType.arrayLength + "]";
-    return getShortenedParamType(paramType.arrayChildren!) + end;
+    return [...getParamType(paramType.arrayChildren!), end];
   } else if (paramType.isTuple()) {
-    return "tuple(...)";
+    if (short) {
+      return ["tuple(...)"];
+    }
+    return [
+      "(",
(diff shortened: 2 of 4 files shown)


wmitsuda:
ok, I think you convinced me, I like it.

but we need to improve the UI, when there is sig + docs + input it becomes too noisy, see the screenshot below.

I think we can add some vertical space before the sig (1 or 2 pixels?).

image

and now I feel like the docs section should have some visual indication that it "belongs" to the sig, what about we try to add a very light yellow bgcolor to the docs box + a left border, like when you add a quote in notion (see a quick and dirty mockup):

image

ah, and rebase please :)

sealer3:
I've added a bit more vertical spacing around the signature.

For the docs background color, amber looks better than yellow in my opinion (`bg-amber-100`):
image

Here's one with a light gray background (`bg-gray-100`):
image

And no background (this looks best to me):
image

Once we get the design settled, I'll rebase. I've added a bit more vertical spacing around the signature.

For the docs background color, amber looks better than yellow in my opinion (`bg-amber-100`):
image

Here's one with a light gray background (`bg-gray-100`):
image

And no background (this looks best to me):
image

Once we get the design settled, I'll rebase. Also, do you want to keep the function docs in the transaction action section or remove it in favor of just one place to view the docs, at this function signature?

sealer3:
Here's with a bit of vertical padding added to the block quote (and with the mono font for the parameters, which I like better as it appears on my machine):
image

wmitsuda:
> Here's with a bit of vertical padding added to the block quote (and with the mono font for the parameters, which I like better as it appears on my machine): image

I think the vertical space + no background looks better.

wmitsuda:
about the font, let's keep the -code style, but I'm curious about mono vs fira code. maybe we can improve the -code style font for everyone, can we move that to another issue? (why it looks better, etc, etc)

sealer3:
@wmitsuda Ready for review, with vertical space + no background in the block quote + keeping `font-code`.

wmitsuda (approved):
lgtm!

wmitsuda:
it is cool that not only it looks cool in the tx details page, but we also got that improvement in our ttrace view "for free"

image

PR #1711, opened 2024-01-19
Add remaining Optimism-specific changes

### LINKED ISSUE #1253: Pull changes from op-otterscan in order to minimize the difference between forks

sealer3:
Primary differences:
* Uses op-erigon specific fields in block and transaction responses
* Adds L1 Security Fees bar to transactions which pay L1 fees
* Adds additional L1 gas/fee details to transactions
* Adjusts block reward calculation to ignore the system transaction in Optimism blocks

New transaction details example (ignore the custom "eETH" symbol which should just be "ETH"):
image

Closes #1253

### DIFF
--- a/src/components/TransactionType.tsx
+++ b/src/components/TransactionType.tsx
@@ -32,6 +32,13 @@ const TransactionType: React.FC<TransactionTypeProps> = ({ type }) => {
         </ExternalLink>
       );
       break;
+    case 126:
+      description = (
+        <ExternalLink href="https://github.com/ethereum-optimism/optimism/blob/develop/specs/deposits.md#the-deposited-transaction-type">
+          Optimism: Deposit
+        </ExternalLink>
+      );
+      break;
     default:
       description = "unknown";
   }
--- a/src/components/TransactionType.tsx
+++ b/src/components/TransactionType.tsx
@@ -34,7 +34,7 @@ const TransactionType: React.FC<TransactionTypeProps> = ({ type }) => {
       break;
     case 126:
       description = (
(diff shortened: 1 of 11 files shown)


src/components/TransactionType.tsx
```diff
@@ -25,6 +25,13 @@ const TransactionType: React.FC<TransactionTypeProps> = ({ type }) => {
         </ExternalLink>
       );
       break;
+    case 126:
+      description = (
+        <ExternalLink href="https://github.com/ethereum-optimism/optimism/blob/develop/specs/deposits.md#the-deposited-transaction-type">
```
wmitsuda, line 30: it seems this link was just moved recently 🙂

sealer3: Updated to https://specs.optimism.io/protocol/deposits.html#the-deposited-transaction-type & rebased

wmitsuda:
could you rebase it plz? it become conflicted after last set of merges.

will give it one more round of testing before merging it.

sealer3:
Done. A quick test shows that it seems to be working as expected.

wmitsuda (approved):

PR #1725, opened 2024-01-26
Show withdrawals on slot page

### LINKED ISSUE #1723: Add withdrawals data to the slot view
Let's update the slot page to add the withdrawals count as well.

image

sealer3:
Closes #1723

I added the withdrawal details to the page:
image

Part of me wants to reduce the vertical padding in each row to make the list more compact.

### DIFF
--- a/src/consensus/slot/Overview.tsx
+++ b/src/consensus/slot/Overview.tsx
@@ -3,16 +3,13 @@ import { FC, useMemo } from "react";
 import { useParams } from "react-router-dom";
 import BlockLink from "../../components/BlockLink";
 import ContentFrame from "../../components/ContentFrame";
-import { balancePreset } from "../../components/FiatValue";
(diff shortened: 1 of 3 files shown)


wmitsuda:
and for some reason I'm getting an exception when pointing it to our mainnet test server (with beacon API from caplin).

maybe it doesn't have the complete data or we found a bug in caplin implementation, could you try to reproduce it?

image

sealer3:
Sure, I will make theses changes. As for the error, I have been following the beacon API specs (see [/eth/v2/beacon/blocks/{block_id}](https://ethereum.github.io/beacon-APIs/#/Beacon/getBlockV2) and click "Schema" to see the Capella and newer example). The `validator_index` should be included in each withdrawal item as a string, like `"1"`.

wmitsuda (approved):
good

PR #1733, opened 2024-01-30
Scroll tables horizontally in small viewport widths

### LINKED ISSUE #291: Block View Text Overflow
I found this bug in the block view from mobile phone.

image

sealer3:
Closes #291.

Uses `table-auto` and a combination of min width and max width classes to enable columns to expand as the viewport width increases. I specifically aimed to prevent column widths from changing when flipping pages.

The PendingResults component is no longer used. We should decide whether that animation should be used in PendingPage or not.

### DIFF
--- a/src/execution/address/GenericTransactionSearchResult.tsx
+++ b/src/execution/address/GenericTransactionSearchResult.tsx
@@ -81,7 +81,7 @@ const GenericTransactionSearchResult = <T extends { hash: string }>({
             total={total}
             totalFormatter={totalFormatter}
           />
-          <StandardScrollableTable isAuto={false}>
+          <StandardScrollableTable>
             {header}
             {items !== undefined ? (
               <StandardSelectionBoundary>
--- a/src/execution/address/GenericTransactionSearchResult.tsx
+++ b/src/execution/address/GenericTransactionSearchResult.tsx
@@ -1,8 +1,8 @@
 import { FC } from "react";
 import ContentFrame from "../../components/ContentFrame";
+import StandardScrollableTable from "../../components/StandardScrollableTable";
 import StandardTBody from "../../components/StandardTBody";
 import StandardTHead from "../../components/StandardTHead";
-import StandardTable from "../../components/StandardTable";
 import { PAGE_SIZE } from "../../params";
 import SearchResultNavBar from "../../search/SearchResultNavBar";
 import { getTotalFormatter } from "../../search/messages";
@@ -81,7 +81,7 @@ const GenericTransactionSearchResult = <T extends { hash: string }>({
             total={total}
             totalFormatter={totalFormatter}
           />
-          <StandardTable>
+          <StandardScrollableTable isAuto={false}>
             {header}
             {items !== undefined ? (
               <StandardSelectionBoundary>
@@ -94,7 +94,7 @@ const GenericTransactionSearchResult = <T extends { hash: string }>({
             ) : (
               <PendingPage rows={PAGE_SIZE} cols={columns} />
             )}
-          </StandardTable>
+          </StandardScrollableTable>
           {total !== undefined && (
             <SearchResultNavBar
               pageNumber={pageNumber}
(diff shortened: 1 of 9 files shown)


src/execution/address/GenericTransactionSearchResult.tsx
```diff
@@ -81,7 +81,7 @@ const GenericTransactionSearchResult = <T extends { hash: string }>({
             total={total}
             totalFormatter={totalFormatter}
           />
-          <StandardTable>
+          <StandardScrollableTable isAuto={false}>
```
wmitsuda, line 84: can we omit the isAuto param here and assume default for a cleaner code?

wmitsuda:
now that we can freely scroll horizontally, I'd suggest we increase the min column size for some columns, I think the defaults don't leave much space for them to be read, see below.

I think besides the Block column, all the other deserves some extra space for better readability.

image

image

wmitsuda (changes requested):

src/execution/address/GenericTransactionSearchResult.tsx
sealer3: Done

sealer3:
The reason for the defaults was to reduce the amount of horizontal scrolling you would need to do on smaller screens, but I see the issue, especially with the method, value, and transaction fee columns.

I increased the column widths and fixed the width of some columns to space the contents better. Let me know if there are any specific columns you would like adjusted. Ideally we would be able to specify a maximum column width for certain columns, but so far I haven't found how to do that.

wmitsuda:
I'd like the tx hash/from/to columns to have the same size as before (see the comparison below), I don't mind if that means the tx fee gets pushed outsize the size + horizontal scrolling.

because they provide the most useful info IMHO in a search results, so we should maximize the ability of the user seeing what they mean (+ provide space for all the auxiliary UI decorations).

I think the other columns are ok. It is better to have horizontal scrolling, but the user can scroll and see more info, than not having scrolling, but the info is truncated.

before:

image


after:

image

sealer3:
I've increased the transaction hash column to match the width of the from/to columns, and I've also made the transaction fee column variable in size but with a small enough default width that it's usually not shown completely, which gives more space to the other columns. Do you think it's okay if the transaction fee column is truncated?

wmitsuda:
I think we can leave the fee be truncated, as it currently is. I'm wondering if we could improve overall UX now that we have given up on trying to fit everything in a fixed resolution :)

but I'd like to think a little about it.

about the last adjustment: tx hash is bigger, but it seems it stole some space from the from/to columns :(

image

sealer3:
I've increased the column width of the from/to columns to compensate. It now takes a width of 1440 px to display all the content.

wmitsuda:
now the tx hash shrank a little 🥲

could you force it to be the same width as before the changes? (which is the hash text without the ellipsis "..." aligned to the "found" word)

image

sealer3:
Okay, the tx, from, and to columns are now expanded to match that (for a clientWidth of 1425 px):
image

wmitsuda:
ugh, got conflicted because I just merged the blob tx PR, rebase plz?

sealer3:
Done

wmitsuda:
I just realized that now the gap between columns are way bigger than before (see below, the before screenshot is taken from sepolia.otterscan.io, you can use the browser inspect tool to compare the gaps more precisely).

I prefer the previous spacing, resulting in more dense info.

before:

image

after:

image

wmitsuda:
I think the other experimental views, like ERC20 search, etc, also have this bigger gap, but that's ok to reduce them as well, I consider it a bug I introduced in the past, i.e., same consistent spacing for all views across the app.

sealer3:
I've reduced the padding between columns, and it now appears to have about the same gap as before. That noticeably helps!

wmitsuda:
yeah, but now it also decreased a few pixels from the margin before the first column.

before:

image

after:

image

sealer3:
Fixed: the horizontal padding around entire table rows is now identical to the way it was before.

wmitsuda (approved):
now we can merge it :)

Issue #1727, opened 2024-01-26
Dencun support


wmitsuda:
This is an umbrella issue to track dencun-related data that we want to show in the UI.

Since our test servers are sepolia, let's wait dencun be activate on sepolia on Jan 30th to better evaluate what to do.

Dencun spec: https://eips.ethereum.org/EIPS/eip-7569

But for now, some interesting data I think we should display:

- EIP-4844 (https://eips.ethereum.org/EIPS/eip-4844) - that'll introduce tx type 3 (blob), we need to ensure it doesn't break UI and proper adjustments are done on info that doesn't make sense for that tx type on transaction details view.
- Also for 4844: block now has blob basefee/gas used/etc.

sealer3:
Tracking ethers support for Dencun properties in blocks: https://github.com/ethers-io/ethers.js/issues/4570

Support for this and Dencun properties in transactions will appear in the next ethers minor release.

PR #1751, opened 2024-02-05
Add Dencun blob tx type

sealer3:
Part of #1727

This PR adds a link to EIP-4844 for type-3 transactions and shows relevant EIP-1559 fields:
image

I've saved the other Dencun work I've done for when we get better ethers support for type-3 transactions, which will appear in its next minor release.

### DIFF
--- a/src/components/TransactionType.tsx
+++ b/src/components/TransactionType.tsx
@@ -25,6 +25,13 @@ const TransactionType: React.FC<TransactionTypeProps> = ({ type }) => {
         </ExternalLink>
       );
       break;
+    case 3:
+      description = (
+        <ExternalLink href="https://eips.ethereum.org/EIPS/eip-4844">
+          EIP-4844 (Blob)
+        </ExternalLink>
+      );
+      break;
     default:
       description = "unknown";
   }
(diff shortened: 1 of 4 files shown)


wmitsuda:
I'm wondering if we it makes sense to show some of the other fields, like input or value.

reading the spec (https://eips.ethereum.org/EIPS/eip-4844#blob-transaction) it seems that they could, in theory, be sent with some meaningful value, so in the context of a block explorer it makes sense to show raw data for inspection.

what if at least we don't try to interpret the 4bytes actions and display "blob" or something instead of the standard "transfer" action in the "Transaction Action" section of tx details page and the "Method" column of tx list? wdyt?

in that case I'd suggest to also use a different background color, i.e., instead of the light yellow (Transfer), maybe light green (Blob)

wmitsuda:
hm, or maybe not, I may have understood incorrectly the semantics of type 3 tx vs regular fields

wmitsuda:
ok, let's ignore my last suggestion, as per conversation in erigon discord, type 3 could be used for execution, so let's not change other fields semantics/display.

PR #1766, opened 2024-02-06
Add error boundary component ErrorFallback

### LINKED ISSUE #1750: Add error boundary for catching unexpected errors
we currently don't have any error boundary configured, so when some exception happens, things just mess up, UI becomes trash and needs to be manually refreshed, no feedback to the user that error may be logged in console, etc...

I think it's time to provide some better error handling by using error boundaries.

reference: https://legacy.reactjs.org/docs/error-boundaries.html

https://www.npmjs.com/package/react-error-boundary

we can iterate on the UI, but I was thinking about some global handler that catches any unexpected error and displays some overlay in the UI with it's content, maybe some instruction to open an issue for us, refresh the page, etc.

it'll be better than the current situation which is: page becomes blank, can't go back, user eventually refreshes the page.

sealer3:
Current look:
image

Because the position of the error boundary is above the react-router-dom outlet, navigation stops working after the error happens, i.e. you can only navigate to the home page. If this is a problem, I suppose we can add an error boundary to each page so the react-router navigation remains functional.

Closes #1750.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,6 +39,7 @@
         "react-chartjs-2": "^5.2.0",
         "react-content-loader": "^6.2.1",
         "react-dom": "^18.2.0",
+        "react-error-boundary": "^4.0.12",
         "react-helmet-async": "^2.0.4",
         "react-image": "^4.1.0",
         "react-intersection-observer": "^9.5.3",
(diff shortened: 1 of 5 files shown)


wmitsuda (approved):
lgtm!

PR #1769, opened 2024-02-08
Show blob icon for blob txs in TransactionLink

### LINKED ISSUE #1768: Signal blob tx on transaction list results
part of dencun support: https://github.com/otterscan/otterscan/issues/1727

we have a space before the transaction hash where we can display some decoration, currently being used only when the transaction fails to display a "revert" icon.

image

I think we could use this same area to display an icon for blob transactions in order to make them easier to identify in tx result lists (criteria: tx type == 3).

I'd like to suggest this initial icon/color: https://fontawesome.com/icons/mound?f=classic&s=solid&pc=%23076ab6

with a tooltip of "Blob transaction"

open to discuss better icon/colors, let's see how it looks like.

sealer3:
Closes #1768

Currently, this uses the `text-purple-400` color as I proposed in #1768, but we can change it to blue:
image

### DIFF
--- a/src/components/TransactionLink.stories.tsx
+++ b/src/components/TransactionLink.stories.tsx
@@ -21,3 +21,10 @@ export const FailedTx: Story = {
     fail: true,
   },
 };
+
+export const BlobTx: Story = {
+  args: {
+    ...SimpleTx.args,
+    blob: true,
+  },
+};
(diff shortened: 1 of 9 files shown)


sealer3:
@wmitsuda Ready for review.

wmitsuda (approved):
finally reached consensus about the UI :)

PR #1804, opened 2024-02-19
Add address tags to address page title and subtitle

### LINKED ISSUE #1650: Add labeled address to page title
An address with a label should have its label appear in the page title. This would make bookmarks and page history easier to manage in the browser.

sealer3:
Closes #1650

Adds a resolved tag string to the page title and also to a "tag" bullet in the address subtitle:
image


wmitsuda (approved):
I'm a little concerned about showing it as a label because we also have another decoration for [ERC20], for example, and it looks like 2 types of labels.

But let's merge it for now, we have some time to think better about it before the next release.

PR #1850, opened 2024-03-10
Tag known superchain addresses according to superchain registry

wmitsuda:
Reference: https://github.com/ethereum-optimism/superchain-registry/tree/main/superchain/configs

Those are L1-side addresses for known L2 chains.


wmitsuda:
@sealer3 review pls

wmitsuda:
only mainnet/sepolia, as op-goerli is being shutdown.

sealer3 (changes requested):
Please make sure all the addresses are checksummed. Non-checksummed addresses do not show up as labels in Otterscan.

Also, I'd suggest against including duplicate labels (Binance). We need to ensure you can distinguish between two separate labels without having to check which addresses they correspond to. For example, if an address receives transfers from two distinct Binance addresses, it should be clear that the addresses are different.

wmitsuda:
fixed the addresses.

regarding the binance one, it was already there, I didn't add it in this PR, but anyway I think it was some random binance addresses I stumbled upon and decided to add it, but really don't know the best approach, i.e., etherscan has binance 1, 2, 3, etc., but that's on whatever criteria they use to discover new ones.

sealer3:
For now, I think we should either choose our own naming convention for this case or use the same labels as Etherscan. Something simple we could do is add an address prefix, such as "Binance: 0x56Ed". We can always change it later.

wmitsuda:
> For now, I think we should either choose our own naming convention for this case or use the same labels as Etherscan. Something simple we could do is add an address prefix, such as "Binance: 0x56Ed". We can always change it later.

I actually like the prefix idea, will retrofit the binance addresses.

sealer3 (approved):

PR #1852, opened 2024-03-10
Add L1 attributes tx addresses

wmitsuda:
Reference: https://specs.optimism.io/protocol/deposits.html#l1-attributes-deposited-transaction

Added for op-mainnet and op-sepolia. For other superchain chains it should be the same, but I didn't add now because (1) I don't have nodes for all of them in order to test (2) we should add more addresses as I'm studying the op-stack specs.

### DIFF
--- a/src/api/address-resolver/hardcoded-addresses/10.json
+++ b/src/api/address-resolver/hardcoded-addresses/10.json
@@ -1,0 +1,4 @@
+{
+  "0xDeaDDEaDDeAdDeAdDEAdDEaddeAddEAdDEAd0001": "L1: Attributes Depositor",
+  "0x4200000000000000000000000000000000000015": "L1: Attributes Predeployed Contract"
+}
--- a/src/api/address-resolver/hardcoded-addresses/10.json
+++ b/src/api/address-resolver/hardcoded-addresses/10.json
@@ -1,4 +1,4 @@
 {
   "0xDeaDDEaDDeAdDeAdDEAdDEaddeAddEAdDEAd0001": "L1: Attributes Depositor",
-  "0x4200000000000000000000000000000000000015": "L1: Attributes Predeployed Contract"
+  "0x4200000000000000000000000000000000000015": "L1: Attributes Predeploy"
 }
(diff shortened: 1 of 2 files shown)


wmitsuda:
@sealer3 wdyt about the names?

sealer3:
I prefer the style of "Optimism: L1 Attributes Contract" and "Optimism: L1 Attributes Depositor" better, since it follows the "organization name: address tag" format. It's a bit tricky since Optimism is the chain itself... otherwise I think "L1 Attributes Contract" and "L1 Attributes Depositor" would also work. I think that the contract has been predeployed can be inferred from the address.

wmitsuda:
> I prefer the style of "Optimism: L1 Attributes Contract" and "Optimism: L1 Attributes Depositor" better, since it follows the "organization name: address tag" format. It's a bit tricky since Optimism is the chain itself... otherwise I think "L1 Attributes Contract" and "L1 Attributes Depositor" would also work. I think that the contract has been predeployed can be inferred from the address.

optimism sounds odd because those are system addresses that occur in any superchain, i.e., Zora has it although Zora's users don't need to know it is an optimism superchain, and even so, "optimism" seems strange, I think in this case "L1" as the prefix is more accurate.

About the predeployed thing, I was in doubt if even the word "contract" would be redundant.  but "L1: Attributes" sounded something was missing in it.

sealer3 (approved):
Ok, sounds reasonable to me.

wmitsuda:
reading the specs again, I think calling them "predeploys" (as in precompiles) would make sense: https://specs.optimism.io/protocol/predeploys.html

so, I think we can call it "L1: Attributes Predeploy"

sealer3:
Since this set of predeployed contracts is common across Optimism networks and is not all related to the layer 1 network, maybe instead we can label them as "Predeploy: L1 Attributes", "Predeploy: L2CrossDomainMessenger", "Predeploy: GasPriceOracle", etc.

Aesthetically, we might also consider using "OP" as a prefix, as in "OP Stack", like "OP: L1 Attributes".

wmitsuda:
I prefer to not use the OP prefix, because in the superchain registry it looks like Optimism is used to refer to the "main" optimism chain, as opposed to Base, Zora, etc.

and predeploy sounds more like a secondary attribute of the contract, I think I'll stick to the L1 prefix (for now).

Issue #245, opened 2022-04-17
feature: custom address labels


totlsota:
This is the number one aspect that keeps me checking other tools/scripts/etherscan. 
With thanks, otterscan is great.

wmitsuda:
You mean, privately attribute names to certain addresses? Sounds good, I just wonder what would be a good UX for persisting this data.

The simplest way would be to store it in browser local storage, but I've heard some time ago Apple would start deleting local storage after some time.

Do you have any idea of a better place you would like this personal data to be stored?

totlsota:
Yes, privately attribute names to addresses: my contracts, other contracts of note. Etherscan has 'tags' and you can add your own 'tags' that will show throughout the UI. For example, I've tagged both the EOA and contract in this example:

<img width="1309" alt="image" src="https://user-images.githubusercontent.com/3116579/163816907-569ad810-22fe-4edc-89a0-0371142f64cb.png">

You can see the default 'tag' from etherscan too [here](https://etherscan.io/address/0x00000000726422a6fECb4759b44D47E48Cf746aa).

<img width="695" alt="image" src="https://user-images.githubusercontent.com/3116579/163817006-faf7b9a4-00f6-4b03-83e6-e0d9d934774d.png">

As a user, I'd be happy to upload a key/value JSON dictionary, or  a command line path to docker even with additional labels. I have no good guidance on where to store user-specific data for this use case "as a web application". The only thing to manage is either a case insensitive map or to convert the addresses to checksum addresses. I looked at the /src/api/address-resolver/ impls and considered just adding to the hard-coded labels manually, but didn't think that was a good plan.

Thanks for considering/ listening ;)

wmitsuda:
Yes, as a little hack, editing the files you pointed out would work.

I think specifying a user provided JSON file would be a good start, if it works for you, although not very user friendly. But I also don't want to block this feature just because I wasn't able to decide on a definitive solution for config persistence.

totlsota:
I wasn't sure the best way to do that when running in docker?

wmitsuda:
You can edit the `1.json` file at `hardcoded-addresses` directory, for example, and rebuild the docker image, where `1` is the chainid for mainnet. This hack should work for now.

But I agree that a proper solution is nice to have.

sealer3:
[...]

PR #1856, opened 2024-03-11
Add initial inline address labeling tool

sealer3:
This tool uses `localStorage` to store address label changes. It has support for loading labels from external lists once a UI is created for that. `localStorage` changes take priority over those from external sources. As it stands, the interface does not let users "clear" labels set by external sources.

Example:
image
image
image

This feature might be useful as is already, but I'm mainly looking for feedback. I'd also like to know how/where to store a list of external label sources (localStorage?) and where the user could manage this. I think we will also need a separate page for managing all of the user's `localStorage` custom labels at once. This might be accessible via an entry in the hamburger button dropdown.

Also, maybe standardizing the circle button into its own component would be useful to do...

Related: #245

### DIFF
--- a/src/api/address-resolver/CustomLabelResolver.ts
+++ b/src/api/address-resolver/CustomLabelResolver.ts
@@ -98,6 +98,15 @@ export class CustomLabelFetcher {
       return undefined;
     }
   }
+
+  public getAllAddresses(): string[] {
+    return Array.from(this.localStorageLabels.keys());
+  }
+
+  public clearAll() {
+    this.localStorageLabels.clear();
+    this.localStorageLabels.clear();
+  }
 }
 
 export class CustomLabelResolver extends BasicAddressResolver {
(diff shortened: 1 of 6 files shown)


wmitsuda (changes requested):
I liked the functionality overall, there are some UI changes I'd like to have in the final version, but I think we don't need to worry about that, because for the prototype I'm happy with a functional interface we can use to test it.

I'd like to go in the following direction from now on, and ask for some small changes:

- Could we introduce some feature flag to enable this extra UI? It could be some WIP_<something> flag, it is only for us, don't need to document/publish it for now. Iit is for us. The reason for that is that we can merge it and keep working on it in small pieces, so we don't need to keep rebasing and dragging an ever growing PR.
- I think a "clear" or "delete" button would be useful to "reset" the state, so we don't need to manually delete the label during our tests.
- yeah, I agree that localStorage is not the ideal solution, it is ok for the purposes of this PR, but from now on we work/discuss the next step of our prototype. I actually have the idea in my mind, I'll share with you offline.

sealer3:
I've hidden the labeler through a feature flag in the config (`WIP_customAddressLabels`) and also added a "clear all" button which wipes all custom address labels from localStorage, for development. To wipe a single address from the UI, edit the name, empty the name string, and click the check mark button to save.

wmitsuda:
note to myself: review this one after release 2.5 which is already full of features.

wmitsuda (approved):

PR #1867, opened 2024-03-12
Stub out calls to useProxyAttributes when ots2 is disabled

### LINKED ISSUE #1854: ots2_getAddressAttributes being called when experimental ots2 flag is disabled
we should not call experimental stuff when it is not enabled. it may have been introduced by myself accidentally when I implemented the experimental features first, but anyway we need to fix it.

in erigon logs:

```
WARN[03-10|21:13:36.636] [rpc] served                             conn=[::1]:54979 method=ots2_getAddressAttributes reqid=20 t=58.605µs err="the method ots2_getAddressAttributes does not exist/is not available"
```

sealer3:
Closes #1854

### DIFF
--- a/src/execution/AddressMainPage.tsx
+++ b/src/execution/AddressMainPage.tsx
@@ -62,7 +62,13 @@ const AddressMainPage: React.FC<AddressMainPageProps> = () => {
     hasCode ? checksummedAddress : undefined,
     provider?._network.chainId,
   );
-  const proxyAttrs = useProxyAttributes(provider, checksummedAddress);
+  const proxyAttrs = config?.experimental
+    ? useProxyAttributes(provider, checksummedAddress)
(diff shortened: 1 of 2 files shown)


wmitsuda (commented):
doesn't this solution violate "no conditional call" rule of react hooks? https://legacy.reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

sealer3:
Yes, it does. The assumption was that the order of hooks would not change, so the conditional would always go one way or the other throughout the page lifecycle. But to follow this rule strictly, I could separate out the experimental tabs and routes into their own components, which we can render conditionally without breaking any of the rules of hooks. I'll do this instead...

wmitsuda (approved):

PR #1883, opened 2024-03-16
Add turn down symbol to Optimism L1 deposit transactions in list

### LINKED ISSUE #1851: Signal "deposited txs" on search results
Like we did for blob txs, on optimism superchains there is a "deposited transaction" (type 126): https://specs.optimism.io/protocol/deposits.html#kinds-of-deposited-transactions

it would be nice if we signal those in the search results.

I'd like to suggest this icon for it: https://fontawesome.com/icons/turn-down?f=classic&s=solid

image

that tx type is used for L1->L2 forced txs but also for a system tx which occurs at position 0 in every block, so it would be easy to test.

sealer3:
Closes #1851 

Example:
image

### DIFF
--- a/src/components/TransactionLink.tsx
+++ b/src/components/TransactionLink.tsx
@@ -33,7 +33,7 @@ const TransactionLink: FC<TransactionLinkProps> = ({
       </span>
     )}
     {deposit && (
-      <span className="text-gray-600" title="Deposit transaction">
+      <span className="text-emerald-600" title="Deposit transaction">
         <FontAwesomeIcon icon={faTurnDown} />
       </span>
     )}
--- a/src/components/TransactionLink.tsx
+++ b/src/components/TransactionLink.tsx
@@ -1,6 +1,7 @@
 import {
   faExclamationCircle,
   faSplotch,
+  faTurnDown,
 } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { FC, memo } from "react";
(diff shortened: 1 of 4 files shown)


src/components/TransactionLink.tsx
```diff
@@ -11,9 +12,15 @@ type TransactionLinkProps = {
...
   blob?: boolean;
+  deposit?: boolean;
 };
 
-const TransactionLink: FC<TransactionLinkProps> = ({ txHash, fail, blob }) => (
+const TransactionLink: FC<TransactionLinkProps> = ({
+  txHash,
+  fail,
+  blob,
+  deposit = false,
```
wmitsuda, line 22: I think it is unnecessary to explicitly define default as `false`, no?

wmitsuda (changes requested):

src/components/TransactionLink.tsx
```diff
@@ -25,6 +32,11 @@ const TransactionLink: FC<TransactionLinkProps> = ({ txHash, fail, blob }) => (
         <FontAwesomeIcon icon={faSplotch} />
       </span>
     )}
+    {deposit && (
+      <span className="text-gray-600" title="Deposit transaction">
```
wmitsuda, line 36: I did test some colors on top of your PR, I think `text-emerald-600` looks better, wdyt?

the gray/black-sh color looks too strong and draw too much attention, IMHO.

image

wmitsuda (commented):
BTW, could you add a storybook story for the deposit case?

src/components/TransactionLink.tsx
sealer3: Sure, I suppose it's better to be explicit.

wmitsuda (approved):

PR #1899, opened 2024-03-20
Add token price resolvers

### LINKED ISSUE #1879: Price feed support on Optimism
We currently only display USD values on mainnet duue to Chainlink only supporting the FeedRegistry contract on mainnet.

That's very annoying, but I really don't want us to hardcode feed addresses on our side. But it is also annoying not having USD info on other chains.

So, I'd like to propose a midterm solution which we can test on Optimism, which is:

1 - Hardcode the ETH/USD price oracle for that chain: https://docs.chain.link/data-feeds/price-feeds/addresses?network=optimism&page=4

2 - Use Uniswap V2 (or V3?) factory `getPair()` to discover/derive other token LPs <-> ETH, and use it to estimate the spot price.

3 - Apply this oracle solution on chainid 10 (optimism mainnet) instead of the feed registry to display USD values when applicable.

sealer3:
Adds support for using UniswapV2 and UniswapV3 spot prices for price estimations, enabling tokens without Chainlink data feeds to have their price estimates shown.

In addition, this change allows the native token USD price data feed contract address to be specified in the config.

Note that the UniswapV3 "is there enough liquidity to trust this price" check is subject to being incorrect if there is a substantial amount of out-of-range liquidity and little or no liquidity in the active range, e.g. if the true price of the token is higher than all active liquidity positions. If it's easy to calculate how much liquidity is on either side of the current token price without having to make too many extra calls, that could be a future improvement.

Other future improvement: Users may wish to know the source of the price data. We may also wish to change the background color of the price estimate based on the source of the information.

Closes #1879.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -99,3 +99,38 @@ Example:
   }
 }
 ```
+
+### Token price oracles
+
+Otterscan makes use of on-chain [Chainlink data feeds](https://docs.chain.link/data-feeds/price-feeds) for fetching historical price data from Chainlink smart contracts compatible with the [AggregatorV3 interface](https://docs.chain.link/data-feeds/api-reference) which act as price oracles. The native token price is fetched from a Chainlink data feed smart contract. The Ethereum mainnet has a registry which Otterscan uses to look up token prices for supported tokens. On other chains, and for tokens not in the registry, Otterscan can estimate a token's price using information from on-chain decentralized exchanges.
+
+**NOTE:** The prices estimated from on-chain decentralized exchanges can be manipulated, especially when tokens have low liquidity or are unable to be traded normally. This price estimation feature returns "best guess" prices; Otterscan certainly cannot guarantee that all tokens with an estimated price have a liquid market or that the prices shown are accurate.
+
+To configure Otterscan's price oracle usage, configure the `priceOracleInfo` key in the config:
+* `nativeTokenPrice`: If configured, shows the native token price at the top of the page.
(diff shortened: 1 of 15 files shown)


wmitsuda:
I think you forgot to embed the configuration for chainId == 10 (op-mainnet), which was the main driver of this issue :)

wmitsuda:
in fact, I was not thinking about adding alternative resolvers on mainnet for now in order to not complicate things, but you rightfully guessed where we were aiming to :)

but yes, regarding L1 mainnet, one of the issues created by having more than 1 resolver is trust and transparency regarding the source. I was thinking about having difference colors for different sources, maybe we change the chainlink to use a "chainlink blue logo"-color and uni use a "uniswap-pink"-color?

sealer3:
Uniswap price sources have a fuschia color. I decided to keep the neutral gray color for Chainlink price sources since I didn't want the price information to become too much of a distraction.
image

For comparison, it would look like this with a blue color:
image


Optimism now has its price oracle information embedded without the need for a custom configuration.

wmitsuda:
> Uniswap price sources have a fuschia color. I decided to keep the neutral gray color for Chainlink price sources since I didn't want the price information to become too much of a distraction. 

ok, let's keep the gray considering Chainlink is the "canonical" oracle. but open to revisit it.

wmitsuda (approved):
finally was able to test it on mainnet, that's a great feature!

PR #1957, opened 2024-04-02
Add Optimism L1 Epoch to block page with optional L1 explorer URL

### LINKED ISSUE #1880: Display block epoch on OP-chains
All superchains have this "epoch" concept, which is the corresponding L1 block a certain L2 block is anchored to.

And since we use archive nodes, that info should be queryable through the L1Block predeploy.

We should display that L1 block reference in the block page, that should reinforce the concept all superchains have their security derived from the L1 chain.

References:

https://specs.optimism.io/protocol/overview.html#epochs-and-the-sequencing-window
https://docs.optimism.io/stack/protocol/smart-contracts#l1block

### LINKED ISSUE #1881: Optionally allow the user to inform an L1 Otterscan URL in an L2 Otterscan instance
This one depends on https://github.com/otterscan/otterscan/issues/1880

Let's say someone is running Otterscan against OP-mainnet, the user should optionally be allowed to inform the URL of his mainnet Otterscan instance and we should cross-reference L1 info, i.e. OP epoch, aka L1 block should link to the L1 Otterscan instance.

That should be useful is the user, for example, has both OP and mainnet running in his dappnode, so he can connect both instances.

I think it could be a static initialization parameter, like the erigon node.

sealer3:
Adds the L1 Epoch fetched from the L1Block contract's "number" function to block pages, and also adds a config option for an L1 block explorer which turns this number into a link.

With an L1 block explorer URL configured:
image

Without an L1 block explorer URL configured:
image

Closes #1880 
Closes #1881

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -134,3 +134,17 @@ Example for Ethereum mainnet:
   }
 }
 ```
+
+### Optimism-specific
+
+For chains which follow the OP Stack, configure the `opChainSettings` key in the config:
+* `l1ExplorerURL`: The root URL of a block explorer for the layer-1 of this chain, without a trailing forward slash. This appears on block pages in the "L1 Epoch" row.
+
+Example for Sepolia:
+```json
+{
+  "opChainSettings": {
+    "l1ExplorerURL": "https://sepolia.otterscan.io"
+  }
+}
+```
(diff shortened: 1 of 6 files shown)


wmitsuda (approved):
lgtm!

Issue #1886, opened 2024-03-17
Solidity code display puts line breaks between lines


sealer3:
E.g., copying the first lines of a verified Uniswap router contract:
```solidity
pragma solidity =0.6.6;



interface IUniswapV2Factory {

    event PairCreated(address indexed token0, address indexed token1, address pair, uint);



    function feeTo() external view returns (address);

    function feeToSetter() external view returns (address);



```

Line breaks should not be doubled.

wmitsuda:
not sure if I got it, does the original verified code have double line breaks?

sealer3:
No it doesn't. Try highlighting and copying the source code from this address to the clipboard and paste it somewhere else: https://sepolia.otterscan.io/address/0xc4ADf4C01DC93Fc54a7d1Fa615B9f1Df56D67132/contract

wmitsuda:
some issue with the highlighter component we are using then?

sealer3:
Yes, I think so. Otterscan has had this issue for at least 1 year. Not being able to copy contracts correctly is not good.

wmitsuda:
regardless of this issue, do you think copy/pasting from the UI is an important feature? because some contracts are spread over multiple files.


if that’s the case, wouldn’t it be more useful to provide a quick way for the user to download the entire set?

sealer3:
Yes, I think it's quite important. But I would say that it's like how being able to copy code directly from the GitHub web UI is useful. You can always run `git clone`, which you're right, we don't have in Otterscan yet, but sometimes you just want to copy a single function or interface.

wmitsuda:
fixed by migrating to Shiki

Issue #1107, opened 2023-07-28
Comments in contract source code create commented line numbers


sealer3:
When there are comments in contract source code, the line numbers also get the commented style (green + italic):
![image](https://github.com/otterscan/otterscan/assets/125761775/ac8b9e22-f308-4498-b569-d4ae6169ac66)

The line numbers should all have the same style.

wmitsuda:
I think this is a bug in the highlighter component itself

wmitsuda:
fixed by migrating to Shiki

PR #1966, opened 2024-04-04
Use Shiki as the syntax highlighter

### LINKED ISSUE #1914: Explore changing the syntax highlighter engine to shiki
I stumbled upon this accidentally while reading vite 5.2 release notes and seeing this shiki thing:

https://twitter.com/vite_js/status/1770455259069968837

it looks more well maintained than the current [problematic] solution we use for syntax highlighting, but needs to be tested.

https://github.com/shikijs/shiki

sealer3:
Uses Shiki instead of react-syntax-highlighter for highlighting both Solidity and JSON (ABIs).

image

Closes #1914, #1107, and #1886.

With Shiki you can copy code to the clipboard correctly. I haven't implemented line numbers yet; apparently that must be done manually. There might be some lag in loading the highlighted component. Shiki does not have much support for React, but at least it's possible following their version 1.0 rewrite.


wmitsuda:
just tried it out and looks excellent so far!

wmitsuda:
I think the decorations capabilities of this editor will allow us to make some interesting things in the future.

sealer3:
Line numbers added:

image

src/execution/address/contract/HighlightedSolidity.tsx
```diff
@@ -0,0 +1,15 @@
+import { FC, lazy } from "react";
+
+import React from "react";
```
wmitsuda, line 3: hm, could you collapse those imports?

src/execution/address/contract/HighlightedSolidity.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+type HighlightedSolidityProps = {
+  source?: string | null;
+};
+
+const HighlightedSolidity: FC<HighlightedSolidityProps> = ({ source }) => {
+  return <HighlightedSource source={source} langName="solidity" />;
+};
+
+export default React.memo(HighlightedSolidity);
```
wmitsuda, line 15: or maybe import `memo` directly, so you don't need to import `React`?

src/execution/address/contract/HighlightedSource.tsx
```diff
@@ -0,0 +1,64 @@
+import { HighlighterCore, getHighlighterCore } from "shiki";
+import langJson from "shiki/langs/json.mjs";
+import langSolidity from "shiki/langs/solidity.mjs";
+import themeGithubLight from "shiki/themes/github-light.mjs";
+import getWasm from "shiki/wasm";
+
+import React, { useEffect, useState } from "react";
```
wmitsuda, line 7: I think it is ok to remove the blank line above and let the import formatter reorg the lines

src/execution/address/contract/HighlightedSource.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+let highlighterSingleton: HighlighterCore | undefined = undefined;
+export const getOrCreateHighlighter = async (): Promise<HighlighterCore> => {
```
wmitsuda, line 10: add a blank line before the export for better readability, please :)

src/execution/address/contract/RawABI.tsx
```diff
@@ -1,16 +1,16 @@
...
-    className="h-60 w-full border font-code text-base"
-    language="json"
-    style={docco}
-    showLineNumbers
-  >
-    {JSON.stringify(abi, null, "  ") ?? ""}
-  </SyntaxHighlighter>
-);
+const RawABI: FC<ABIAwareComponentProps> = ({ abi }) => {
+  return (
```
wmitsuda, line 6: I think this code can be simplified by removing the "{ return {" since the is no code, only a tag, like the one who is being replaced.

wmitsuda (changes requested):

wmitsuda (approved):

PR #1984, opened 2024-04-11
Add e2e tests with Anvil backend

### LINKED ISSUE #1853: Add anvil support in our integration tests
Like we have an erigon devnet in our integration tests, it would be helpful to have it run against an anvil node.

Although the anvil support was not implemented by us, it would be useful to check if we didn't break any call which would be caught only when testing it against anvil (for whatever reason) and/or to check it against anvil versions (for fail fast and fast feedback to them).

sealer3:
Closes #1853 

Currently, an anvil binary from a fixed recent nightly version is downloaded and used for the backend.

### DIFF
--- a/.github/workflows/end-to-end.yaml
+++ b/.github/workflows/end-to-end.yaml
@@ -110,3 +110,49 @@ jobs:
         with:
           name: Cypress Devnet Screenshots
           path: cypress/screenshots
+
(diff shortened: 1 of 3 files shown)


wmitsuda (approved):

wmitsuda:
let's merge and pray it'll work :)

PR #2001, opened 2024-04-14
Support function return values in traces

### LINKED ISSUE #1530: Add return value support to traces if the contract is verified
This is a current limitation of our trace view, since we were relying only on 4bytes matching until recently. The 4bytes db has only signatures with parameters, not return value.

But if the contract is verified on Sourcify we have the entire ABI, hence we can reliably parse the return value.

I don't remember if the custom API Otterscan depends on already returns the return data, otherwise we'll need to add that in Erigon first.

sealer3:
Closes #1530

image

The PR includes devdoc support for return values, and the same decoder is applied to the Read Contract page.

### DIFF
--- a/src/execution/address/contract/ReadContract.tsx
+++ b/src/execution/address/contract/ReadContract.tsx
@@ -60,6 +60,11 @@ const ReadContract: React.FC<ContractsProps> = ({
                     <ReadFunction
                       func={FunctionFragment.from(fn)}
                       address={checksummedAddress}
+                      devMethod={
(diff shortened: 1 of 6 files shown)


wmitsuda (changes requested):
it looks great, just a small UI adjustment: could we make the return box take all horizontal space when expanded like we do when the inputs are expanded?

see the following screenshot with both input and return boxes expanded for comparison:

image

wmitsuda (approved):

PR #2022, opened 2024-04-21
Use consensus constants from fetched consensus spec

### LINKED ISSUE #2011: Add support to CL explorer on gnosis
Our CL explorer works ok-ish on gnosis chain, but there was differences in the beacon chain specs that messes everything in the UI, i.e. smaller epoch, block times, etc.

see: https://github.com/gnosischain/specs/blob/master/beacon_chain.md

also their consensus specs: https://github.com/gnosischain/specs/blob/master/consensus/config/gnosis.yaml

as a side note, I think that consensus spec file format is based on ETH parameters, I wonder if it would be a good solution to embed and read this format so any other chain with different params can be easily be added, but it is up to you.

sealer3:
Closes #2011 

Fixes Gnosis chain timestamp

### DIFF
--- a/src/consensus/epoch/LoadingSlotItem.tsx
+++ b/src/consensus/epoch/LoadingSlotItem.tsx
@@ -1,13 +1,13 @@
 import { FC, memo } from "react";
 import ContentLoader from "react-content-loader";
-import { slot2Epoch, useProposerMap } from "../../useConsensus";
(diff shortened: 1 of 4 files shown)


wmitsuda (approved):
that's good, I didn't realize the config was queriable from the API itself.

PR #2045, opened 2024-04-25
Only run Pages deployment on pushes to develop

### LINKED ISSUE #2016: `pages-deployment` workflow failures
* The `on: [push]` in the integration-deployment workflow means that the workflow will run on pushes to any branch
* Without a check that the secrets exist, this workflow will fail on Otterscan forks which don't have the secrets set up.

I recommend adding a filter so that this workflow runs only on pushes to the develop branch, and adding a secrets check to the integration-deployment workflow so that the jobs are skipped if the requisite secrets are not set.

sealer3:
Closes #2016.

I found no way around getting "workflow invalid" errors using conditional statements, so Otterscan forks will have to disable this workflow manually. But still, without this change, Otterscan contributors will always get a "workflow invalid" notification on every push by default. That is why this PR is necessary, and I think it reflects what we want anyway.

### DIFF
--- a/.github/workflows/integration-deployment.yaml
+++ b/.github/workflows/integration-deployment.yaml
@@ -2,6 +2,12 @@ on:
   push:
     branches:
       - develop
+  pull_request:
+    types:
+      - opened
+      - synchronize
+  workflow_dispatch:
+
 jobs:
   deploy-mainnet:
     uses: ./.github/workflows/pages-deployment.yaml
--- a/.github/workflows/integration-deployment.yaml
+++ b/.github/workflows/integration-deployment.yaml
@@ -9,8 +9,24 @@ on:
   workflow_dispatch:
 
 jobs:
+  check-secrets:
+    runs-on: ubuntu-latest
+    outputs:
+      has-secrets: ${{ steps.has-secrets.outputs.defined }}
+    steps:
+      - name: Check if deployment secrets are available
+        id: has-secrets
+        shell: bash
+        run: |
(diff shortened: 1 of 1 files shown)


wmitsuda:
which error is this "workflow invalid"?

anyway, some reasoning to give context:

actually, I do want to run in other branches, for example, when some PR gets pushed we get those nice "preview" like deploys.

we used to get those "for free" on Cloudflare when we used their standard workflow (CF pulls from git), but there is a limitation (from CF side) which they only allow 1 CF project per git repo.

that's why I changed it and inverted the deploy (according to their own recommendtion), now 1 git repo pushes to N CF pages projects using their "direct upload" feature.

one thing that we may fine tune (not sure yet) is maybe disable it for dependbot PRs.

sealer3:
Pushing commits to a fork, I get this notification:

image

The provided reason the workflow is invalid is that there is no deployment permission given to the workflow:

image

What is frustrating is that even making the job conditional does not make this error go away, and the entire workflow file is considered "invalid".

wmitsuda:
ok, let's then add the secret checking you suggested before and close this one, would that work?

sealer3:
Unfortunately, a secrets check is not enough in this case. GitHub frustratingly provides no way to check or set permissions from within a job, since permissions are evaluated in the "Set up job" step, before if statements are evaluated, according to https://github.com/orgs/community/discussions/63736.

We'll have to change the workflow triggers instead. https://github.com/cloudflare/pages-action/issues/94 explains this:

> One commonly used workaround is to use labels as triggers.
> Since labels can only be granted with Write permission to the repository, this is considered a reasonable workaround

With this solution, we can add a label to PR branches that we want to run deployments on.

sealer3:
I think I got it!

The issue was that `on: [push]` is somehow too broad. The PR pushes are correctly captured with `pull_request`. Secret checking now seems to work.

wmitsuda (approved):

wmitsuda:
hope everything continue working. but we have a lot of time to fix CI until the next release :)

PR #2092, opened 2024-05-15
Add experimental support for GitHub address label backend

sealer3:
Adds a settings page at `/settings` where GitHub credentials (username/repo name/API token) can be set. No requests are made to GitHub unless all three are configured. This page is accessible from a button in what is currently the Sourcify options menu.

image

If GitHub credentials are configured, Otterscan will fetch and update labels from the target repository.

I plan to add the "local storage only" backend as an option in the settings page.


wmitsuda:
I didn't forget about this one :)

but now I'd like to first implement SSR, then visit this one, I think it'll make things simpler.

PR #2104, opened 2024-05-17
Add Docker env var for entire Otterscan config

### LINKED ISSUE #2074: Feature Request - expose 'assetsURLPrefix' as an environmental variable for Docker
Per the title - it would be great to expose this parameter for users of Docker

sealer3:
Closes #2074 via `OTTERSCAN_CONFIG` env var. Also adds `ASSETS_URL_PREFIX` for the assets URL and `DISABLE_CONFIG_OVERWRITE` if the user wants to mount the config file manually.


wmitsuda (approved):
ok, agree

PR #2105, opened 2024-05-19
Scroll to log fragment and add log links

### LINKED ISSUE #2102: Add anchors to log indexes
just saw this on twitter: https://x.com/0xtarc/status/1790742671046164933

Sounds a good suggestion.

sealer3:
Closes #2102.

Adds links to each log circle and scrolls to a particular log index if a fragment is found in the page location.

There is a timeout because log components do not notify their parents that they have finished loading (i.e., expanded completely) unless we build out infrastructure for calling back to the parent, which I think is overkill for this feature. For a similar reason, we have to implement the scrolling logic in the Logs component because a particular log index element is created dynamically and does not exist at page load. We can switch from smooth to instant scrolling if there is a preference for that.

### DIFF
--- a/src/execution/transaction/Logs.tsx
+++ b/src/execution/transaction/Logs.tsx
@@ -9,11 +9,11 @@ type LogsProps = {
 };
 
 const Logs: FC<LogsProps> = ({ logs }) => {
-  let location = useLocation();
+  const location = useLocation();
   useEffect(() => {
     setTimeout(() => {
       if (location.hash) {
-        // Scroll to fragment, e.g. "#log-3"
+        // Scroll to fragment, e.g. "#3"
         let foundElement = document.getElementById(location.hash.slice(1));
         if (foundElement) {
           foundElement.scrollIntoView({
--- a/src/execution/transaction/Logs.tsx
+++ b/src/execution/transaction/Logs.tsx
@@ -1,5 +1,6 @@
 import { Log } from "ethers";
-import { FC, memo } from "react";
+import { FC, memo, useEffect } from "react";
+import { useLocation } from "react-router-dom";
 import ContentFrame from "../../components/ContentFrame";
 import LogEntry from "./LogEntry";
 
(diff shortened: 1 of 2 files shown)


src/execution/transaction/log/LogIndex.tsx
```diff
@@ -1,13 +1,18 @@
 import { FC } from "react";
+import { Link } from "react-router-dom";
 
 type LogIndexProps = {
   idx: number;
 };
 
 const LogIndex: FC<LogIndexProps> = ({ idx }) => (
-  <span className="flex h-12 w-12 items-center justify-center rounded-full bg-emerald-50 text-emerald-500">
+  <Link
```
wmitsuda, line 9: is it possible to use just numbers for anchor id?

so we can have a nicer url like:

http://blablabla/tx/logs#353

instead of

http://blablabla/tx/logs#log-353

as it is implied the /logs main content is a list of logs (and I don't think we'll add other stuff in there)

src/execution/transaction/Logs.tsx
```diff
@@ -1,28 +1,45 @@
...
-          </>
-        ) : (
-          <div className="py-4 text-sm">Transaction didn't emit any logs</div>
-        )}
-      </>
-    )}
-  </ContentFrame>
-);
+const Logs: FC<LogsProps> = ({ logs }) => {
+  let location = useLocation();
```
wmitsuda, line 12: could you change it to `const` instead of `let` to keep it as the standard?

src/execution/transaction/log/LogIndex.tsx
```diff
@@ -1,13 +1,18 @@
...
+    to={"#log-" + idx}
+    className="flex h-12 w-12 items-center justify-center rounded-full bg-emerald-50 text-emerald-500"
```
wmitsuda, line 11: what about adding an underline style on hover to make it explicit now the logs number ball is linkable?

wmitsuda (changes requested):

wmitsuda (approved):

PR #2115, opened 2024-05-22
Add raw transaction broadcast page

### LINKED ISSUE #1965: Add broadcast transaction feature
Other block explorers let you broadcast a signed transaction from the explorer interface. It would be useful to have this in Otterscan, too.

sealer3:
Closes #1965

This page is accessible from `/broadcastTx`. There is no link to it from elsewhere in Otterscan. I think we can add it as a menu item to what is currently the Sourcify source dropdown.

Example:
image


wmitsuda:
I think it is ok to add the menu item in this PR.

sealer3:
This is what it looks like:
image

PR #2146, opened 2024-05-30
Add dark mode

### LINKED ISSUE #280: Dark mode
Is there any way to enable dark mode?

sealer3:
Dark mode is enabled when the browser's color scheme preference is set to dark.

With color inversion + hue rotation providing most of the dark colors, it should be easy to maintain. One-off changes can be set with the `dark:` class prefix.

image

Closes #280.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" class="dark:bg-white dark-mode">
   <head>
     <meta charset="utf-8" />
     <link rel="icon" href="/favicon.ico" />
--- a/src/Footer.tsx
+++ b/src/Footer.tsx
@@ -1,5 +1,4 @@
 import React, { useContext } from "react";
-import ThemeToggler from "./components/ThemeToggler";
 import { RuntimeContext } from "./useRuntime";
 
 const Footer: React.FC = () => {
@@ -7,9 +6,6 @@ const Footer: React.FC = () => {
 
   return (
     <>
-      <div className="bg-white">
-        <ThemeToggler />
-      </div>
       <div
         className={`w-full border-t border-t-gray-100 px-2 py-1 text-xs ${
           provider?._network.chainId === 1n
--- a/src/Footer.tsx
+++ b/src/Footer.tsx
@@ -8,8 +8,8 @@ const Footer: React.FC = () => {
     <div
       className={`w-full border-t border-t-gray-100 px-2 py-1 text-xs ${
         provider?._network.chainId === 1n
-          ? "bg-link-blue text-gray-200"
-          : "bg-orange-400 text-white"
+          ? "bg-link-blue text-gray-200 dark:text-gray-900"
+          : "bg-orange-400 text-white dark:text-black"
       } text-center`}
     >
       {provider ? (
--- a/src/Footer.tsx
+++ b/src/Footer.tsx
@@ -1,23 +1,29 @@
 import React, { useContext } from "react";
+import ThemeToggler from "./components/ThemeToggler";
 import { RuntimeContext } from "./useRuntime";
 
 const Footer: React.FC = () => {
   const { provider, config } = useContext(RuntimeContext);
 
   return (
-    <div
-      className={`w-full border-t border-t-gray-100 px-2 py-1 text-xs ${
-        provider?._network.chainId === 1n
-          ? "bg-link-blue dark:bg-link-blue-light text-gray-200 dark:text-gray-800"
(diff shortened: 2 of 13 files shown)


wmitsuda:
interesting solution, looks good overall.

it follows system settings, should we add a toggle to force light/dark as well?

wmitsuda (changes requested):
I think we should change the bottom blue bar, it is too bright in dark mode.

image

sealer3:
Oh, there is still no dark toggle yet...

wmitsuda:
> Oh, there is still no dark toggle yet...

I think it should be System/Light/Dark, with system as default like in other websites.

sealer3:
I added a theme toggler that appears in the bottom-left corner in the footer:

image

Note that it has its own background color:

image

The theme is saved in local storage, following Tailwind's documentation for the most part. If a user's system dark mode preferences changes, Otterscan will automatically adapt to it.

wmitsuda:
I have some mixed feelings about the toggler:

- The "sun" and the "gear" icons look similar if you don't pay enough attention (I think it is more the style of that specific "sun" is to blame though...).
  - if we don't find a "better" sun, maybe try a lightbulb? maybe regular lightbulb vs crossed lightbulb?
- I think it draws too much attention in the home page, sure the user should be able to change the setting from there, I just think the whole toggle is too big and too distracting if you don't want to care about it.
  - Also in the regular pages, it feels it doesn't belong to where it is, i.e., it is between the page content and the footer, but it is not a footer neither page content. In this case I would hide it under the sandwich menu top-right.
- What if we hide it in the sandwich menu top-right and make it (only the menu, not the entire header) appear in the home page? WDYT?

wmitsuda:
re: local storage, it seems fine to me

sealer3:
I think we can do that - hide it in the menu and add the menu to the home page.

sealer3:
Here's the new look:
image

And opened on the home page:
image

The app context is not available on the home page since it is housed in the `<Main>` element which includes the header, so the sourcify source option is hidden on the home page right now. We could potentially refactor it to give the app context to the home page.

wmitsuda:
I changed my mind :)

WDYT?

image

image

I think what I didn't like in the first attempt was the solid icons being too distracting, using the regular ones look better. Also using a display icon instead of the gear because the gear also feels like it is going to open a settings dialog.

```typescript
import { faMoon, faSun } from "@fortawesome/free-regular-svg-icons";
import { faDisplay } from "@fortawesome/free-solid-svg-icons";
```

it also feel more like the menu in tailwindcss website:

image

wmitsuda:
the rounded icons still look too distracting and "heavy", I think we can turn them into standard menu items with icons, like in the tailwindcss site

sealer3:
I think switching to regular icons is fine, and I do like the sun/moon icons better. I think using standard menu items would look best if the theme buttons are in their own menu, like on the Tailwind website.

sealer3:
What about this? I reduced the padding, reduced the border, and faded the background color even more:
image

wmitsuda:
> What about this? I reduced the padding, reduced the border, and faded the background color even more: 

still don't like it, it looks busy (thick border + rounded corners + background color) and doesn't feel like it belongs the menu it is contained in (that "a component from another app got imported as-is here"-feeling)

sealer3:
Using standard menu items and title:
image

Then I moved Broadcast Transaction to the bottom:
image

wmitsuda:
I think we don't need to repeat the word "theme" since it has its own section now and call the items "Light/Dark/System".

And make it evident that the "Light/Dark/System" items are hierarchically contained in the "Theme" section, I feel like the current UI language doesn't feel like that because the "Theme" font is smaller. I think we can keep the font smaller (it is saying to the user that label is not important), but there is something missing (no ideas yet, maybe a different shade of gray/black?).

Also, make the "Data source" and "Theme" texts non-selectable, they are not "text".

sealer3:
What about a slight left margin to convey their hierarchical status as children of the groups?
image

wmitsuda:
I think it looks better without the left margin.

And I think only "System" instead of "System default" feels better (like in tailwind website). Less is more :)

wmitsuda:
Small glitch I found: the hamburger menu looks a little smaller (vertical) in the home page vs regular pages, probably due to be anchored to the height of search box.

Can we make it the same height in the home to be pixel perfect?

sealer3:
Fixed to use "System". And at most zoom levels, the height should now be identical.

wmitsuda (approved):
ok, I think it is good for merging, we can make small adjustments separately.

PR #2148, opened 2024-05-31
Add slot and validator columns to Blocks Rewarded table

### LINKED ISSUE #1356: Show validators who produced blocks in fee recipient blocks list
It's useful to see which validators produced blocks whose fee recipient was set to a particular address, such as when multiple validators operate in a pool.

- [x] Add a column to the fee recipient blocks produced list for the validator
- [x] Ensure that the list still renders without this information when the beacon RPC is not available

sealer3:
Closes #1356 

The beacon block is found from the parent beacon block root in the execution block header.

Example:
image

When old slot data (or parent beacon block root) is not available:
image

If the beacon API is not set in the config, these new beacon chain columns will not be shown.

### DIFF
--- a/src/execution/address/BlockRewardedItem.tsx
+++ b/src/execution/address/BlockRewardedItem.tsx
@@ -1,9 +1,15 @@
-import { FC, memo } from "react";
+import { FC, memo, useContext } from "react";
 import BlockLink from "../../components/BlockLink";
 import NativeTokenAmount from "../../components/NativeTokenAmount";
 import TimestampAge from "../../components/TimestampAge";
+import SlotLink from "../../consensus/components/SlotLink";
+import ValidatorLink from "../../consensus/components/ValidatorLink";
 import { BlockNumberContext } from "../../useBlockTagContext";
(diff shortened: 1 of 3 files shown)


wmitsuda:
does it work it there is a missed slot after the one being displayed? i.e.: is the parent beacon block root of the next block == current slot if there is a missed slot between current block and next?

sealer3:
Yes. According to [EIP-4788](https://eips.ethereum.org/EIPS/eip-4788) "Even in the event of missed slots since the previous block root does not change, we only need a constant amount of space to represent this “oracle” in each execution block"

So the parent beacon block root is for the last non-missed beacon block, since missed slots do not change the parent beacon block root. Here's an example: https://beaconcha.in/slot/9248711. Now slot 9248710 was missed. Notice that the parent root link takes you to slot 9248709.

You can verify that block 20043698 (beacon block 9248709) shows the validator and slot correctly for fee recipient 0x88c6C46EBf353A52Bdbab708c23D0c81dAA8134A in Otterscan:
image

Here's the slot page:
image

PR #2153, opened 2024-06-01
Show address transaction count for EOAs

### LINKED ISSUE #744: Account transaction count indicator
Etherscan displays the number of transactions an account has had above an address's transactions table. This is useful because it tells you how much activity there is. At present, you cannot tell in Otterscan if an address had 30 transactions or 30,000 transactions without going into a recent transaction and looking at the nonce.

So, we should find something to display above the table, maybe to the left of the pagination buttons, that gives an indication of the amount of activity the account has had. One example of this for EOAs might be the account's current transaction count, which is easy to query. Another might be the total number of incoming and outgoing transactions, which is more useful for contracts, which don't send any transactions.

sealer3:
Closes #744 

Adds an address's transaction count next to the page transaction count. It's only shown for EOAs.

image

There's possibly other text that would fit here, such as using "X outgoing transactions" instead.

Maybe even "X total outgoing transactions"...

### DIFF
--- a/src/execution/address/AddressTransactionResults.tsx
+++ b/src/execution/address/AddressTransactionResults.tsx
@@ -16,10 +16,16 @@ import { useFeeToggler } from "../../search/useFeeToggler";
 import StandardSelectionBoundary from "../../selection/StandardSelectionBoundary";
 import { ProcessedTransaction } from "../../types";
 import { BlockNumberContext } from "../../useBlockTagContext";
-import { useAddressBalance, useContractCreator } from "../../useErigonHooks";
+import {
+  useAddressBalance,
+  useContractCreator,
+  useHasCode,
+  useTransactionCount,
+} from "../../useErigonHooks";
 import { useResolvedAddress } from "../../useResolvedAddresses";
 import { RuntimeContext } from "../../useRuntime";
 import { usePageTitle } from "../../useTitle";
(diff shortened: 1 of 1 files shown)


wmitsuda:
I feel like this info should be in the previous section, below or at the right of the balance box.

it doesn't mix well with the pagination stuff, it feels in the wrong place.

wmitsuda (changes requested):
let's try to improve this info placement

sealer3:
Okay. I don't think it deserves its own line below Balance - and we would have to stop it from page jumping if we put it below anyway. It's not easy to come up with other places where it would fit... the closest I'm thinking of is some kind of nonce-like indicator at the subtitle line. What do you think about that? Imagine a green nonce box (with the up arrow) to the right of the address label.

The reason I had it next to the pagination is that the transaction count gives you a lower bound on the total number of transactions in the table.

wmitsuda:
put it on the right? make it half and half panel

sealer3:
> put it on the right? make it half and half panel

Exactly half?
image


Or thirds?
image

wmitsuda:
half looks better?

wmitsuda:
that somehow still feels a little odd, but the feature is good. I think we'll eventually make changes in the layout in the future.

wmitsuda (approved):

PR #2171, opened 2024-06-07
Add previous day price change to price box

### LINKED ISSUE #1039: Show 24-hour price change near the current native token price
It's valuable to know how the current price of the native token (usually ETH) is doing compared to yesterday. Near the current price of ETH in the header, we should show the relative change in price from 24 hours ago. To obtain the older price, we can just perform the price call on top of a block number from 24 hours ago.

- [ ] Show relative change in price formatted as (+1.23%) or (-4.56%)
- [ ] Mark a positive change in green and a negative change in red

sealer3:
Closes #1039 

Adds the price change since 24 hours ago to the price box:
image

A binary search finds the exact block number from 24 hours ago. A more efficient way of finding it (like an indexer) would be an improvement.


wmitsuda (approved):
look nice, let's try it

PR #2209, opened 2024-06-16
Dependabot/npm and yarn/headlessui/react 2.0.4

sealer3:
Closes #2125.

Still needs to be tested, but it builds correctly.


sealer3:
After testing everything appears to be functional. The only thing I could not test was the camera scanner.

wmitsuda (approved):

PR #2237, opened 2024-06-23
Add whatsabi ABI filling prototype

### LINKED ISSUE #2018: Add support for ABI guessing
we should take a look at: https://github.com/shazow/whatsabi

it could be useful for non-verified contracts.

it looks like it can work totally standalone, just providing the contract code, but we should be careful and take a deeper look at their code to make sure we are not introducing any external dependencies/calls.

sealer3:
Here's a preview:
image

You can use the Read Contract tab, but the return values aren't decoded due to the function selector not using it and therefore 4bytes not having it either: 
image

Function selectors not found in 4bytes aren't displayed currently. With some more lifting we could include them.

- [x] Do a more comprehensive review of the whatsabi repo.

Closes #2018

### DIFF
--- a/src/execution/address/Contracts.tsx
+++ b/src/execution/address/Contracts.tsx
@@ -53,7 +53,7 @@ const Contracts: React.FC<ContractsProps> = ({ checksummedAddress, match }) => {
             <FontAwesomeIcon icon={faWarning} className="mr-2" />
             <span className="text-md">
               Contract not found in Sourcify respository. Below is an estimate
-              of the ABI from its bytecode:
+              of the ABI from its bytecode.
             </span>
           </div>
         </>
--- a/src/execution/address/Contracts.tsx
+++ b/src/execution/address/Contracts.tsx
@@ -53,7 +53,7 @@ const Contracts: React.FC<ContractsProps> = ({ checksummedAddress, match }) => {
             <FontAwesomeIcon icon={faWarning} className="mr-2" />
             <span className="text-md">
               Contract not found in Sourcify respository. Below is an estimate
-              of the ABI from its bytecode using known function selectors:
+              of the ABI from its bytecode:
             </span>
           </div>
         </>
@@ -102,7 +102,10 @@ const Contracts: React.FC<ContractsProps> = ({ checksummedAddress, match }) => {
         {match !== undefined && match !== null && (
           <>
             {match.metadata.output.abi && (
-              <ContractABI abi={match.metadata.output.abi} />
+              <ContractABI
+                abi={match.metadata.output.abi}
+                unknownSelectors={match.unknownSelectors}
+              />
             )}
             {match.type !== MatchType.WHATSABI_GUESS && (
               <div>
--- a/src/execution/address/Contracts.tsx
+++ b/src/execution/address/Contracts.tsx
@@ -1,4 +1,4 @@
-import { faChevronDown } from "@fortawesome/free-solid-svg-icons";
+import { faChevronDown, faWarning } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { Menu, MenuButton, MenuItem, MenuItems } from "@headlessui/react";
 import React, { lazy, useContext, useEffect, useState } from "react";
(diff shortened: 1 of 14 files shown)


sealer3:
Whatsabi seems to have minimal dependencies (only ethers), and I do not see any immediate risks to using it. It makes use of `node:fs` in its examples but does not refer to code that reads from the filesystem except in one of the debug tools.

I updated the code to show the unknown function signatures to give a more complete picture of the contract:
image

I do not sort the unknown function selectors, but we might consider sorting them and the functions in the ABI.

wmitsuda:
I gave it a try and now I have mixed feelings:

- Nowadays most contracts seems to be verified on Sourcify (which is good)
- The ones who are not, they usually appear as "unknown selectors" for me (mainnet), which defeats totally the purpose of this feature. In fact, I've been browsing mainnet for 1/2 hour, randomly trying to find a non-verified contract that returns me the signatures, but all my attempts are "unknown".

So, maybe it is better to not implement this one, WDYT?

sealer3:
There was a mistake in the 4bytes URL format (no 0x prefix should be used). I fixed this and also made the list of known functions sorted. Please try it again.

wmitsuda:
ah, now I see it correctly.

src/execution/address/Contracts.tsx
```diff
@@ -44,7 +44,21 @@ const Contracts: React.FC<ContractsProps> = ({ checksummedAddress, match }) => {
...
+      {match && match.type === MatchType.WHATSABI_GUESS && (
+        <>
+          <div
+            className="bg-orange-100 border-l-4 border-orange-500 text-orange-700 px-3 py-2.5 mb-4 rounded mt-2"
+            role="alert"
+          >
+            <FontAwesomeIcon icon={faWarning} className="mr-2" />
+            <span className="text-md">
+              Contract not found in Sourcify respository. Below is an estimate
+              of the ABI from its bytecode:
```
wmitsuda, line 56: I think we can replace the ":" by a ".", since this is a warning inside a box, there no continuity from the warning to the ABI tree, so a ":" looks a little odd here.

src/sourcify/useSourcify.ts
```diff
@@ -219,10 +284,28 @@ export const useSourcifyMetadata = (
...
-      sourcifyFetcher([...key, sourcifySources]),
+    fetcher,
+  );
+  if (error) {
+    return null;
+  }
+  return data;
+};
+
+export const useWhatsabiMetadata = (
```
wmitsuda, line 298: does it make sense to separate whatsabi-related code into it own `.ts` file? so we don't mix Sourcify and whatsabi stuff and the code becomes more modular and clear in what are their connections.

wmitsuda (changes requested):

src/sourcify/useSourcify.ts
sealer3: I agree.

wmitsuda:
> I do not sort the unknown function selectors, but we might consider sorting them and the functions in the ABI. Another thing to consider is that we don't know the state mutability of the functions. I set them all to `view` so you can try Read Contract on them.

ok, I think we can proceed to properly integrate it, I liked the result of the prototype so far, so let's make it a proper feature.

Having them on read contract is useful; yes it is impossible to know the return (if any) encoding or if they are mutable functions.

Some changes I think we should do from now:

1 - Add the same warning box in the "Read contract" tab so the user knows it is a guessed ABI
2 - Should we assume the return value is a blob of bytes and show the standard "Raw" textbox we show for `[]byte` so the user can at least examine and take their own conclusions about what is returned?

wmitsuda:
note to myself: I'm using the contract `0xB86E490E72F050c424383d514362Dc61DaBB1Cc3` on mainnet for testing because it has a good mix of unknown functions, known views and clearly non-view functions.

src/sourcify/useSourcify.ts
sealer3: I put the new file in the `sourcify` directory, since it shares many of the same types and functionality as `useSourcify`. Feel free to suggest an alternative.

wmitsuda: we can move it later if we find a better organization

sealer3:
We cannot have ethers interpret the return value as `bytes1[]` or similar because the contract ABI specification requires dynamic types to have the data length (in words) before the data. I've just disabled the Decoded tab for these return values.

wmitsuda:
> We cannot have ethers interpret the return value as `bytes1[]` or similar because the contract ABI specification requires dynamic types to have the data length (in words) before the data. I've just disabled the Decoded tab for these return values.

ah yes, that's what I meant to say, just the raw. and it is good that you left the "Decoded" button disabled instead of removing it because it clearly communicates the user the decoding is not available.

wmitsuda (approved):

PR #2258, opened 2024-06-29
Add call options accordion to Read Contract functions

### LINKED ISSUE #2243: Allow historical querying on "Read Contract" tab
Currently the "Read Contract" tab allows eth_calling methods based on chain tip.

Because we rely on archive nodes, we could allow the user to optionally call it in the context of a previous block:

- That would allow one to easily compare results over time.
- That would allow one to call methods on self-destructed contracts (by setting the block to sometime before it was destroyed).

I think be default everything should assume "latest" block and we should add some "not-showy" textbox where the user can optionally enter a block number.

UI/UX challenges:

- A contract may have a lot of methods, should we show a box after each "Query" button? will the UI be too cluttered?
- Or just 1 textbox after the beginning for simplicity and cleanness, but potentially make the user scroll to the top to change the value?
- In theory a contract (pre-dencun and who knows in other chains...) could be destroyed and resurrected many times, so we should have some nice error messages so the user knows some call is failing because the contract was not yet deployed at that time, or it was destroyed and not yet resurrected.

sealer3:
Closes #2243 

Adds an accordion controlled by a button with extra call options: block number, sender, and call value.

When closed:
image

When open:
image

A note is appended to the error message if the the block number was changed and contract address has no code.

### DIFF
--- a/cypress/e2e/devnet/verified-contracts.cy.ts
+++ b/cypress/e2e/devnet/verified-contracts.cy.ts
@@ -18,39 +18,51 @@ describe("Read Contract tests", () => {
           .contains("getVariableLengthStringArray")
           .parent()
           .as("func");
-        cy.get("@func").find("input").type("one");
+        cy.get("@func").find("input.w-full").type("one");
         cy.get("@func").find("button").contains("Add Element").click();
-        cy.get("@func").find("input").eq(1).type("two");
+        cy.get("@func").find("input.w-full").eq(1).type("two");
         cy.get("@func").find("button").contains("Add Element").click();
-        cy.get("@func").find("input").eq(2).type("three");
+        cy.get("@func").find("input.w-full").eq(2).type("three");
 
-        cy.get("@func").find("input").eq(0).should("have.value", "one");
-        cy.get("@func").find("input").eq(1).should("have.value", "two");
-        cy.get("@func").find("input").eq(2).should("have.value", "three");
+        cy.get("@func").find("input.w-full").eq(0).should("have.value", "one");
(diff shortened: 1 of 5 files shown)


wmitsuda:
Some tweaks that I think we should do at first glance:

- The arrow down button after the function name makes the whole UI feel busy, I'd suggest moving it to right of the "Query" button.
  - Also, move the entire block/sender/value section to after the query button.
  - I think that would make this entire section feels like an optional group of settings. Maybe changing the "arrow button" to something else would improve this feeling (maybe "+" meaning "extra call options"?)
- I think the textbox with the arrows up/down to increment/decrement the number is unnecessary, given the magnitude of block number is of millions on no-ghost chains. It is faster for the user to just delete the characters and change the number.
  - We can for sure reduce the horizontal space for block number field by a half.
    - Same for value field, although I think "value" can have a bigger width than "block number".
  - We could add an auxiliary button just after the block number textbox to set it to the max (latest) block number as an aid.

sealer3:
Here is what it looks like with the suggested changes:
image

wmitsuda:
@sealer3 sorry, totally forgot about this one, could you rebase it pls?

sealer3:
@wmitsuda Done

wmitsuda:
I thought about framing those fields under some "call options" group, but that would make the UI very noisy. I think using colors may be enough to signal that's an "extra group of fields which are not function parameters".

sealer3:
New look (dark/light):

image
image

wmitsuda (approved):
looks good.

PR #2261, opened 2024-06-29
Add ids and scrolling to Read Contract functions

### LINKED ISSUE #2256: Add anchor links to functions in the "read contract" tab
Each method should have its own anchor (like we did in the logs view), so user can link directly to it.

I suggest the anchor to be the method selector string (unless we have a better idea).

### LINKED ISSUE #2257: Link view methods in the contract tab to their equivalents in the read contract tab
Depends on: https://github.com/otterscan/otterscan/issues/2256

The contract tab offers a good outline of what is available in the contract, but I often see some interesting method and need to switch tabs to invoke it in the read tab.

That turns to be annoying in big contracts because you need to switch tabs and scroll to the corresponding method form, too many clicks.

We should linkify the entries in the ABI view to their read form sections in the read tab.

image

sealer3:
Closes #2256
Closes #2257 

Note that there are no links on the React Contract page itself, but the functions can be linked to. @wmitsuda Do you want the numbers in the ordered list to be links?

### DIFF
--- a/src/execution/address/Contracts.tsx
+++ b/src/execution/address/Contracts.tsx
@@ -96,6 +96,7 @@ const Contracts: React.FC<ContractsProps> = ({ checksummedAddress, match }) => {
               <ContractABI
                 abi={match.metadata.output.abi}
                 unknownSelectors={match.unknownSelectors}
+                address={checksummedAddress}
               />
             )}
             {match.type !== MatchType.WHATSABI_GUESS && (
--- a/src/execution/address/contract/ContractABI.tsx
+++ b/src/execution/address/contract/ContractABI.tsx
@@ -6,7 +6,11 @@ import { ABIAwareComponentProps } from "../../types";
 import DecodedABI from "./DecodedABI";
 import RawABI from "./RawABI";
 
-const ContractABI: FC<ABIAwareComponentProps> = ({ abi, unknownSelectors }) => (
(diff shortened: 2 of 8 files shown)


sealer3:
These are the links in the ABI to the Read Contract page:
image

The links are currently unstyled. I'd appreciate some suggestions on how they should look.

wmitsuda:
could you rebase this one as well? for some reason even checking out this code and running against erigon 3 is broken (it is working with erigon 2). not sure if there is something buggy in erigon 3 or something in otterscan which was fixed recently.

wmitsuda:
anyway, I think we can linkify the entire function - return string, e.g.:

"function blabla() view returns (address)"

wmitsuda:
I think we can style it with underline on hover + light blue text, but that's a guess, let's see how it looks like

sealer3:
> could you rebase this one as well? for some reason even checking out this code and running against erigon 3 is broken (it is working with erigon 2). not sure if there is something buggy in erigon 3 or something in otterscan which was fixed recently.

This was fixed in a recent commit. Several Otterscan users were reporting issues with this, and switching from `latest` to `develop` fixed their problems.

wmitsuda (changes requested):
just the styling pending

sealer3:
@wmitsuda Done!

wmitsuda (approved):

Issue #2286, opened 2024-07-10
Infinite config reloading after QR scan


sealer3:
I've noticed an "infinite config reloading" issue and have found a way to reproduce it reliably. To reproduce:

1. Make a QR code with the following text: 0x0000000000000000000000000000000000000000
Here's one:
![image](https://github.com/otterscan/otterscan/assets/125761775/2142414d-3dae-49ef-b9c7-9593cd27a853)
2. Scan the QR code using the QR reader.
3. Otterscan will redirect to http://localhost:5173/search?q=0x0000000000000000000000000000000000000000, resulting in the Page Not Found page.
4. Notice in the console that the config is being reloaded in an infinite loop, and that your webcam is still activated, with the QR reader somehow still running in the background.

This issue is certainly related to, and may be due to, the recent config loader or router changes, since I only began observing infinite config reloads after the router changes were merged to the develop branch. I have only tested this on a vite dev server.

This is related to #2259.

wmitsuda:
actually, I figured the QR scanner seems to be broken for a while, it gets into this loop in very old versions, which is even more strange.

wmitsuda:
trying to track down when in the code it broke.

wmitsuda:
apparently I deleted the /search Route on commit 33939054f2bc6435b19452da62f2deac8cac66e8

maybe I deleted it temporarily in order to migrate to react-router v6, but forgot to restore it (?)

really don't remember, let me try to restore it.

wmitsuda:
but anyway, there are 2 issues here:

- camera support dropped (needs to be restored)
- infinite rendering loop (it was happening already before the recent migration)

wmitsuda:
reopening because it is not entirely fixed.

Issue #2259, opened 2024-06-29
Nonce navigation results in infinite config reload loop


sealer3:
Navigation by nonce now results in Otterscan reloading the config infinitely. For example, try

http://localhost:5173/address/0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266?nonce=0

PR #2319, opened 2024-07-16
Fix missing search route

wmitsuda:
that should fix https://github.com/otterscan/otterscan/issues/2286

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,6 @@
         "@fortawesome/react-fontawesome": "^0.2.2",
         "@headlessui/react": "^2.1.2",
         "@otterscan/react-qr-reader": "^5.2.0",
-        "@rollup/rollup-linux-x64-gnu": "^4.18.1",
         "@shazow/whatsabi": "^0.12.0",
         "@storybook/test": "^8.1.11",
         "@testing-library/jest-dom": "^6.4.6",
@@ -46,6 +45,7 @@
         "react-router-dom": "^6.24.1",
         "shiki": "^1.10.3",
         "swr": "^2.2.2",
+        "tiny-invariant": "^1.3.3",
         "use-keyboard-shortcut": "^1.1.6",
         "web-vitals": "^3.5.2"
       },
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "react-router-dom": "^6.24.1",
     "shiki": "^1.10.3",
     "swr": "^2.2.2",
+    "tiny-invariant": "^1.3.3",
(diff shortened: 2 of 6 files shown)


wmitsuda:
actually, this one fixes the route, but not the refreshing loop, for some reason earlier today it was not happening anymore, but now it is...

sealer3 (changes requested):
Thanks for fixing the route. This PR doesn't fix the infinite config reloading issue however, so I don't think it should close #2286 when merged. Could you check whether the infinite config reloading happens on your end? I've tried multiple browsers, and they exhibit the same behavior.

wmitsuda:
yes, it doesn't fix the reloading, will rework the PR :)

wmitsuda:
The latest commit should fix the case when it gets an error because it is updating a parent state, probably because of navigate() function in-sync with the UI thread.

However I think the infinite loop is not related, I tried to put a log inside the `evaluateScan` function and for the case when the reload loop still happens, the log happens as well.

So what I believe is happening in this case is:

- camera does not get closed and scanning (can confirm that from the "scan" log being continously emited)
- the event keeps being emited, but with the already scanned code
- the redir keeps being executed

that explains the infinite config reloading, because it is like the same QR code being infinitely being scanned and redirected.

I'd recommend we merge this PR as is, as all fixes I did here are valid regardless of the camera problem and we continue discussing the camera issue in https://github.com/otterscan/otterscan/issues/2320

sealer3 (approved):
Okay, I think we can merge this. The infinite config reloading bug still needs to be addressed, but I don't think it's because of the camera.

There are other outstanding problems (they occur on the develop branch too): when you open the QR scanner, you see the ETH node checking page again briefly. Also, when I issue console logs in the QR scanner, they appear in groups of two. So it seems that all the components are duplicated somehow.

Also, #2259 results in the same infinite config reloading behavior, probably also due to routing.

wmitsuda:
I think #2259 can be solved by a similar solution to this one, will send a PR soon.

re: duplicate log, I'm almost sure it is bc: https://stackoverflow.com/questions/72489140/react-18-strict-mode-causing-component-to-render-twice

you can mitigate the double log confusion by installing react devtools extension: https://www.frontendundefined.com/posts/react-issues/double-console-react-18/

> when you open the QR scanner, you see the ETH node checking page again briefly.

hm, let me try to see this one

sealer3:
You're right, half the logs were in a lighter color. I didn't understand why the `scan` log was appearing in groups of two, but maybe it's for the same reason.

PR #2349, opened 2024-07-22
Add autofix

wmitsuda:
I'd like to test: https://autofix.ci/


wmitsuda:
note: I think we could possibly merge this action with the existing build+prettier one, but I left it separated just in case we try it/don't like it and want to remove it.

we can maybe merge both actions in ~2 weeks?

sealer3 (approved):
Sounds good!

PR #2350, opened 2024-07-22
Use transition for redir

wmitsuda:
fixes https://github.com/otterscan/otterscan/issues/2259

I've been running this for a couple of days and haven't noticed the loop anymore.

I think the previous code was already wrong and the router refactor only made the issue evident.


sealer3:
I no longer get the infinite config reloading, which is great, but I still see the config loading logs appear (`Using hardcoded config`, etc.) from the `loadOtterscanConfig` function, which print once. Do you know why this happens?

wmitsuda:
that i’ll investigate next, will do a separate PR

sealer3 (approved):

PR #2360, opened 2024-07-24
Fix config reload on nonce navigation

wmitsuda:
- Moved config/runtime load to page level.
- For some reason the RouterProvider gets rerendered when navigating sometimes causing the reload; decoupling the load from page render solves that.
- I was planning to do this decoupling anyway regardless of this bug.


sealer3 (approved):
Looks good

PR #2381, opened 2024-07-29
Increase log scroll timeout to 500ms

### LINKED ISSUE #2376: Sometimes the log anchor does not position correctly
Minor glitch, very low priority, but when I refresh the page (e.g. below) multiple times, sometimes it positions the screen at the input table, not at the beginning of the log entry.

http://localhost:5100/tx/0xce349e2454621c7c4ce34e5cf94b46c8208a06d4d96425e1c5166db5266fec5b/logs#555

sealer3:
Closes #2376.

I am able to reproduce the issue 35% of the time (20 attempts) with the following network connection (using browser debug tools):

Download = 450 kbps
Upload = 150 kbps
Latency = 150 ms

The issue stems from the fact that the page (and each component) is not aware of when all the logs are completely loaded and their height fully extended, so we have to guess when the logs are likely to be loaded and then scroll. Further, we don't want to scroll the page too late after it finishes loading. I'd rather not complicate the code too much, so I just extended the timeout from 200ms to 500ms. I did not get any log scroll misses in another 20 attempts with this 500ms timeout.

We could consider making it an "instant scroll" rather than an animated scroll, though that wouldn't change whether the logs are still loading.

### DIFF
--- a/src/execution/transaction/Logs.tsx
+++ b/src/execution/transaction/Logs.tsx
@@ -21,7 +21,7 @@ const Logs: FC<LogsProps> = ({ logs }) => {
           });
         }
       }
-    }, 200);
+    }, 500);
   }, [logs]);
   return (
     <ContentFrame tabs>


wmitsuda (changes requested):
yeah, I actually think the scrolling effect sounds a little hacky :)

have you investigated why this is actually required? i.e. why the browser can't simply position the page at the correct anchor place by itself?

sealer3:
On a page load event, the browser computes the exact position of the anchor and jumps there. On our page, the page load event occurs after all the essential assets load but well before the target anchor's component loads, and we also have an asynchronous request to fetch the log signatures (4bytes/Sourcify). When these log signature requests resolve, the heights of the log components increase to render each log and parameters. If the page scrolls before all the logs are expanded to their full height, it will not scroll far enough for when all logs above the target one finally load.

We can fix this by (a) delaying the scroll until after the data has loaded somehow or (b) fixing the heights of the log components to some predefined value - not ideal. There might be some other approaches for (a) which I haven't considered yet.

wmitsuda:
I see... I think we can try https://reactrouter.com/en/main/components/scroll-restoration , but that requires moving data into loaders (which you are already working on).

maybe we can hold on this issue until we refactor the rest of the code?

sealer3:
I think holding off on this until after loaders are implemented is a good idea. We'll have to test whether ScrollRestoration scrolls to hashes. I don't see them mentioned on that page.

PR #2383, opened 2024-07-29
Wrap block/position and from/nonce details

### LINKED ISSUE #2380: Block/position row on tx details is breaking on iPad
See screenshot below.

At very first we should force the individual elements to not break on whitespace, then we need to think about how to break the bigger elements.

Maybe remove the vertical divider "|" and stack both elements vertically in this case?

image

sealer3:
Closes #2380.

Allows the Block/Position and From/Nonce transaction detail rows to wrap.

Enough room:
image

Compressed:
image

### DIFF
--- a/src/components/BlockConfirmations.tsx
+++ b/src/components/BlockConfirmations.tsx
@@ -6,7 +6,7 @@ type BlockConfirmationsProps = {
 };
 
 const BlockConfirmations: FC<BlockConfirmationsProps> = ({ confirmations }) => (
-  <span className="rounded bg-gray-100 px-2 py-1 text-xs text-gray-500">
+  <span className="rounded bg-gray-100 px-2 py-1 text-xs text-gray-500 whitespace-nowrap">
     {commify(confirmations)} Block{" "}
     {confirmations === 1 ? "Confirmation" : "Confirmations"}
   </span>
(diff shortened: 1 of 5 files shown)


wmitsuda (changes requested):
hm, not sure if I like the solution, instead of simply allowing wrapping, I think we should:

- hide that vertical divider, which looks ugly when the second block gets wrapped.
- give some more vertical space between the wrapped blocks
- both blocks should be left aligned

wmitsuda:
also, as seen in the original issue screenshots, I think we should (independently of wrapping of not) make the individual components more resilient by:

- the "<<", ">>" navigation buttons should be made "not wrappable" regardless of their size
- the same for the "N confirmations" and "N / N" boxes

of course that would not solve this issue, it would actually make the entire row overflow or wrap by itself in this case, but I think it is worth it to apply those fixes as well.

sealer3:
I have implemented everything listed except hiding the vertical divider when it is part of the first element in a new flex row. I spent a while trying to find an alternative that doesn't resort to using media queries for the viewport width, but there is no easy way of doing it. If it's any consolation, other top block explorers also have the vertical divider separating items between rows.

New vertical spacing example:
image

wmitsuda:
> I have implemented everything listed except hiding the vertical divider when it is part of the first element in a new flex row. I spent a while trying to find an alternative that doesn't resort to using media queries for the viewport width, but there is no easy way of doing it. If it's any consolation, other top block explorers also have the vertical divider separating items between rows.
> 
> New vertical spacing example: image

they are all ugly IMHO :)

but let's merge it, we can change it later in case of better ideas

Issue #2388, opened 2024-07-29
[epic] Refactor current architecture


wmitsuda:
Currently, the entire app is built on top of fetch-on-render using ugly `useEffect`s mixed with `useSWR`s.

We should explore more the data loading APIs from react router in order to move to a fetch-in-the-router strategy + Suspense on pendind data: https://reactrouter.com/en/main/guides/data-libs

That will bring us several benefits:

- Much cleaner code.
- Parallel fetching (for now while we are still 100% SPA).
- Moving to react-router data loading APIs would make it an incremental step to a bigger move to remix/react-router 7 -> SSR based architecture.

This is an umbrella issue for 1 of N PRs.

I'd suggest we incrementally start refactoring the `/address/*` routes to this model since the code there is becoming spaghetti, and the routes with nested panels seems to fit exactly the main case react-router + data loading aims to solve.

We can use it as an exercise to see what works and what don't (although I've already migrated the top-level App router to use data loaders) and evaluate if everything keeps working.

Also we should consider incrementally moving from SWR to React Query, because:

- SWR seems not very active, IIRC we stopped upgrading it because of some typing bug that was not getting attention.
- SWR fetching is tied to the hooks model, while React Query provides both hook and not hook support for fetch APIs.

PR #2390, opened 2024-08-01
Add data loader to address page

sealer3:
For #2388

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -90,9 +90,18 @@ const loader: LoaderFunction = async () => {
 };
 
 const addressLoader: LoaderFunction = async ({ params, request }) => {
-  const rt = await runtime;
   return defer({
-    hasCode: rt.provider.send("ots_hasCode", [params.addressOrName, "latest"]),
+    hasCode: runtime.then((rt) =>
+      rt.provider.send("ots_hasCode", [params.addressOrName, "latest"]),
+    ),
+  });
+};
+
+const addressTxResultsLoader: LoaderFunction = async ({ params }) => {
+  return defer({
+    balance: runtime.then((rt) =>
+      rt.provider.getBalance(params.addressOrName ?? ""),
+    ),
   });
 };
 
@@ -158,7 +167,11 @@ const router = createBrowserRouter(
           element={<Address />}
           loader={addressLoader}
         >
-          <Route index element={<AddressTransactionResults />} />
+          <Route
+            index
+            element={<AddressTransactionResults />}
+            loader={addressTxResultsLoader}
+          />
           <Route
             path="txs/:direction"
             element={<AddressTransactionResults />}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -18,6 +18,7 @@ import ConnectionErrorPanel from "./ConnectionErrorPanel";
 import Footer from "./Footer";
 import Home from "./Home";
 import Main from "./Main";
+import { PAGE_SIZE } from "./params";
 import ProbeErrorHandler from "./ProbeErrorHandler";
 import { loader as searchLoader } from "./Search";
 import { ConnectionStatus } from "./types";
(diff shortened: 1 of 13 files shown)


sealer3:
This is a first attempt at getting AddressMainPage to use a data loader. It fetches whether the address has code right when navigation begins.

@wmitsuda This change causes the whole page to flash when you change tabs. Do you have any idea why this happens?

sealer3:
The reason I saw the tabs flashing is that the address tabs are now lazy-loaded. This greatly reduces the Address page size. Previously it was the largest page code to load at 1,095 kB, and now it is 39.56 kB. Nonetheless we might want to consider finding a way to load some of the tab code early.

sealer3:
Sadly there is still a full content shift when the address changes. I'm not sure why, since the loader's return values are completely deferred now.

wmitsuda:
> The reason I saw the tabs flashing is that the address tabs are now lazy-loaded. This greatly reduces the Address page size. Previously it was the largest page code to load at 1,095 kB, and now it is 39.56 kB. Nonetheless we might want to consider finding a way to load some of the tab code early.

do you mean, the code imports? I think would be ok to not-lazy load some of them or apply some load tricks (I have some in mind if you want to go that way).

sealer3:
Yes, that's what I meant, the code for the tabs are lazy-loaded. I think it's great that the address page from a cold start loads faster but think it would be best if all the tab code eventually loads on its own. If that doesn't seem possible we can just make the Contract tab and any other frequently used ones non-lazy.

I would really like to hear if you have any ideas for why there is a full content shift when the address changes. I haven't yet figured it out yet.

wmitsuda:
> Yes, that's what I meant, the code for the tabs are lazy-loaded. I think it's great that the address page from a cold start loads faster but think it would be best if all the tab code eventually loads on its own. If that doesn't seem possible we can just make the Contract tab and any other frequently used ones non-lazy, as you said.

I think you can try the technique described here (strategy 3, lazy load + trigger on tab hover): https://mikeguoynes.medium.com/the-ultimate-guide-to-react-lazy-loading-4cacd8bd3cf0

> 
> I would really like to hear if you have any ideas for why there is a full content shift when the address changes. I haven't yet figured it out yet.

by address change you mean it flashes only at the first tab change, right? probably because the lazy tab suspends, after that I can't see any flashes while switching tabs

sealer3:
No, I mean when you click different addresses in the transaction results, there is a flash every time on my end.

wmitsuda:
ah, I see... I suspect all the chain of useEffects while rendering the AddressMainPage is leading to the flash on rerendering, maybe the checksummedAddress calculation.

I think moving those useEffects to loaders will eliminate most of those flashes, or at least cause the Suspends on more narrow parts of the page.

wmitsuda:
the code for useAddressOrENS is really ugly :(

what if we decouple the checksummed address fixer from it, so it does not block the rendering of the page, and the calculation of an eventual reverse ENS name can be progressively rendered?

wmitsuda:
(those are my guesses looking quickly at the code, I may be guessing wrong)

sealer3:
After stubbing it out, useAddressOrENS doesn't seem to be the cause. What I've noticed is that immediately (t = 0ms) after clicking a new address, useContractCreator and useTransactionCount from AddressTransactionResults are fired, using the old address. This happens also on a production build. I will keep investigating.

sealer3:
`useLoaderData` is the culprit of the re-renders. Apparently it is a common complaint that the expected behavior is to re-render like this: https://github.com/remix-run/react-router/issues/10756 https://github.com/remix-run/react-router/issues/11134 https://github.com/remix-run/react-router/issues/11477

The solution according to the maintainers is to speed up renders:

> This is expected behavior. When you navigate, React Router updates useNavigation().state from idle to loading and remains on the current page while the loaders run for the next page. This causes the current page to re-render so that it can show any pending/optimistic UI based on useNavigation().
> 
> Re-renders are at the heart of React and are usually not an issue unless you have expensive/slow components, and you can optimize them individually if they are problematic: https://kentcdodds.com/blog/fix-the-slow-render-before-you-fix-the-re-render

There is some discussion of how to get around a re-render (by **not** using defer) here: https://github.com/remix-run/react-router/issues/10816

sealer3:
Rerenders are not the root cause of the issue. Increasing the latency of network requests to the Erigon RPC server, which is used by the data loader, increases the length of the "flash" during which the entire AddressMainPage isn't shown. This happens regardless of whether we use `defer`.

wmitsuda:
So, it is an "expected" flashing because it is loading new data? 

Maybe we can minimize that effect by using something like useDeferredData to keep rendering the old data until the new one is available?

wmitsuda:
I guess maybe the previous code didn't have the flashing because we are somehow (I don't remember) already rendering the old data until it is replaced by the new, avoiding the flashing.

sealer3:
Wrapping `useDeferredValue` around `useLoaderData` seems to work!

wmitsuda (approved):

PR #2425, opened 2024-08-13
Remove checks assuming provider object is now guaranteed to be present

wmitsuda:
Now that we guarantee that the provider object is created at initialization, we can remove lots of checks downstream and simplify the code.


sealer3 (approved):
I removed a few more checks that I found.

wmitsuda:
uh thanks, for some reason my search didn't caught them all

PR #2470, opened 2024-09-17
Erigon 3 removed total difficult field

wmitsuda:
Current builds of E3 removed the "total difficult" field, so there may be E2 or old E3 nodes with that info, others without it.

Current Otterscan builds have the block page broken with recent E3 nodes.

### DIFF
--- a/src/execution/BlockDetails.tsx
+++ b/src/execution/BlockDetails.tsx
@@ -167,7 +167,9 @@ const BlockDetails: FC<BlockDetailsProps> = ({ blockNumberOrHash }) => {
             {commify(block.difficulty.toString())}
           </InfoRow>
           <InfoRow title="Total Difficulty">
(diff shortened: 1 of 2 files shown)


sealer3:
Is Erigon breaking the Ethereum Execution API spec with this change, or is the API spec also changing?

And is total difficulty undefined for recent blocks only or for all block responses? (If it's for all of them, I think we should hide the entire row in that case.)

wmitsuda:
yeah, the spec changed: https://github.com/ethereum/execution-apis/pull/570

it should be undefined for all responses once the client software is upgraded. I'd like to leave it "N/A" on purpose to explicitly communicate this was an existing field which was removed, at least for now to not surprise any users with information disappearing between releases.

We can revisit this one and omit it in the future though.

sealer3 (approved):
Agree, thanks for linking to that PR.

PR #2529, opened 2024-09-29
Handle 404 and 5xx beacon responses in SlotNotFound

### LINKED ISSUE #2498: Properly handle missed slots
Currently we display an error-ish page when:

- Slot is missed
- CL doesn't have that data (because node is pruned, didn't backfill, etc.)

image

I think we can now try to search for a more elegant solution, if it exists and it is not expensive (i.e. have to do a lot of networks calls to CL API), in order to detect the correct scenario and display a nicer page to the user:

- If the slot is missed, we could display the standard slot page linked to the expected validator who missed it and blank fields like we do in the epoch page.
- If we could reliably CL doesn't have data, we could explain to the user he needs to enabled archive data in CL, backfill, etc...

sealer3:
Distinguishes beacon node errors, 404 responses, and unknown errors from one another.

Closes #2498

### DIFF
--- a/src/consensus/slot/Overview.tsx
+++ b/src/consensus/slot/Overview.tsx
@@ -51,7 +51,7 @@ const Overview: FC = () => {
       {isLoading ? (
         <OverviewSkeleton />
       ) : error ? (
-        <SlotNotFound slot={slotAsNumber} />
+        <SlotNotFound slot={slotAsNumber} error={error} />
       ) : (
         <>
           <InfoRow title="Timestamp">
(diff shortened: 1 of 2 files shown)


sealer3:
To check whether a slot was missed, this PR will query the next beacon block header (next slot). If it exists, it will check if its parent comes before the slot in question. If so, the slot was definitely missed. If two slots in a row are missed (rare on mainnet), the first one will show the generic 404 response.

`/slot/10070963` (slot not found):
image

HTTP 500 response:
image

Data otherwise not available (404 response):
image

wmitsuda:
I'm seeing some strange behavior in my node, let me first clarify the node is working as expected in order to avoid validating this PR against buggy responses :)

https://github.com/erigontech/erigon/issues/12209

wmitsuda (approved):

PR #2557, opened 2024-10-10
Show whatsabi selectors when assetsURLPrefix is not configured

### LINKED ISSUE #2441: How to get WhatsAbi working?
Context: Working on running [otterscan for local anvil instance](https://github.com/0xMetacrypt/otterscan-devnet) but running into issues.

I'm unable to get WhatsAbi to show _anywhere_, not on Contract pages, not on transactions and I'm not sure why. Tried looking at env variables but there doesn't seem to be an enable/disable flag for it.

WhatsAbi would be very useful since sourcify no longer works (ipns 500s on gateways, and central repo is errors on CORS) and loading selectors from 4byte is a great way to get _some_ idea on the transaction and traces page.

sealer3:
Fixes #2441

### DIFF
--- a/.github/workflows/end-to-end.yaml
+++ b/.github/workflows/end-to-end.yaml
@@ -38,7 +38,7 @@ jobs:
       - name: Set up xvfb
         run: |
           sudo apt update
-          sudo apt install -y libgtk2.0-0 libgtk-3-0 libgbm-dev libnotify-dev libnss3 libxss1 libasound2 libxtst6 xauth xvfb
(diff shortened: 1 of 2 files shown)


sealer3:
Failing CI checks are due to a Cypress browser dependency not existing as a package in the Ubuntu image due to an upgrade from Ubuntu 22.04 -> Ubuntu 24.04 in the Ubuntu latest image.

sealer3:
I've included the updated `apt install` list, though it won't be reflected until this is merged. Please review the workflow change, too.

wmitsuda (approved):

Issue #2554, opened 2024-10-05
Otterscan RPC API: return whether internal calls succeeded in trace


sealer3:
In the trace tab, we try to decode revert messages using the expected type even if the internal call reverts. See `/tx/0x578235d94f187fac181d929a1d43b7e094834d30c422b64d608871dea60d68dc/trace` - the decoded return value is `true` but actually, if you look at the raw output data of that call, it's a revert message.

So `ots_traceTransaction` should include in its response whether each internal call succeeded or not. Even transactions which succeed may have an internal call which reverts.

We should show all internal reverts with the same red X we use for reverted transactions.

This change is necessary for an efficient implementation of #2500, namely to find where the revert reason actually originated if it bubbles up to the outer call. It is also necessary for analyzing such cases in the trace if the revert reason changes on its way up.

CanutoIX:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

PR #2579, opened 2024-10-29
Show revert locations in contract source code

### LINKED ISSUE #2499: Track, follow and provide feedback on Sourcify API v2
We currently don't integrate with the Sourcify API directly, but we should on the future, more specifically the API v2 which is on design phase.

Resources: 

https://sourcify.stoplight.io/docs/sourcify-apiv2/branches/main/
https://github.com/ethereum/sourcify/issues/1470
https://github.com/ethereum/sourcify/issues/1545

I think we should keep supporting the offline data formats which is very convenient for researchers, but also supporting the API would be better for fresh data.

### LINKED ISSUE #2500: Point to contract code location where transaction failed
With SourcifyV2 (filesystem-based Sourcify sources could also support this) we can request the `sourceMap` in the response. We can take the trace of a failed transaction to see where in the bytecode a transaction or internal transaction failed, and then point the user to the exact location in the source code that caused it.

Might require
* Access to trace API

### LINKED ISSUE #3185: Use Sourcify APIv2 for contract lookups
A while ago, we launched the Sourcify APIv2 that leverages the DB to provide information-rich responses. We are moving away from being a filesystem based service to a DB based one.

With that, we are also deprecating the legacy api that was based on the filesystem format e.g. https://repo.sourcify.dev/contracts/partial_match/369/0xd07cF2Dec7303cB140BB8514739eA549E1E9d394/metadata.json 

Checking our traffic, we can see a good number of such requests come from Otterscan instances (particularly for PulseChain). Currently we handle these by redirecting such requests from https://repo.sourcify.dev to the server's `/repository` endpoint so

1. https://repo.sourcify.dev/contracts/partial_match/369/0xd07cF2Dec7303cB140BB8514739eA549E1E9d394/metadata.json
2. Redirected to: https://sourcify.dev/server/repository/contracts/partial_match/369/0xd07cF2Dec7303cB140BB8514739eA549E1E9d394/metadata.json
3. Served by the handler actually based on the DB data, that formats everything to conform the legacy API.

As a primary consumer of the Sourcify data, we would love to see Otterscan to use the APIv2 and move away from our legacy API. The APIv2 also provides a lot of interesting data that you can leverage to provide a better view into a verified contract. You can check the blog post about [v2 lookup endpoints](https://docs.sourcify.dev/blog/apiv2-lookup-endpoints/) or check https://repo.sourcify.dev that uses these v2 endpoints.

The code that consumes the metadata.json through IPFS data can remain the same, we are still hosting exact matches and files on IPFS. 

For the "Sourcify Servers" datasource, APIv2 would be the way to go. Let me know what you think and if we can help anywhere.

sealer3:
On the transaction page of any reverting transaction, a trace is done on the transaction to find the program counter offsets in the contract bytecode of each internal call. Otterscan then determines the "revert chain", maps bytecode to instructions for contracts with Sourcify source mappings, and using the source mappings provides links to the source code, highlighting the specific Solidity statements which caused the revert. Out-of-gas and runtime error (e.g., division by zero) locations are not currently supported since we currently check for a final REVERT opcode in the trace to signal an internal transaction failure. Implementation of #2554 would enable us to extend this to those other cases.

The revert locations are offered for all calls in the revert chain, since the outer transaction might not have an interesting revert reason, and the desired origin might be deeper in the call stack.

Todos:
- [x] Scroll to first highlighted statement automatically
- [x] Need suggestions for design of revert chain display
- [x] Requires SourcifyV2 support, which is not released yet
- [x] Add option in config

Closes #2500.
Closes #2499.
Closes #3185.

### DIFF
--- a/src/execution/address/ContractFromRepo.tsx
+++ b/src/execution/address/ContractFromRepo.tsx
@@ -1,6 +1,7 @@
 import { faCircleNotch } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import React from "react";
+import { DecorationOptions } from "shiki";
 import { MatchType, useContract } from "../../sourcify/useSourcify";
 import { useAppConfigContext } from "../../useAppConfig";
 import HighlightedSource from "./contract/HighlightedSource";
@@ -12,6 +13,7 @@ type ContractFromRepoProps = {
   fileHash: string;
   type: MatchType;
   langName: string;
+  decorations?: DecorationOptions["decorations"];
 };
 
 const ContractFromRepo: React.FC<ContractFromRepoProps> = ({
@@ -21,6 +23,7 @@ const ContractFromRepo: React.FC<ContractFromRepoProps> = ({
   fileHash,
   type,
   langName,
+  decorations,
 }) => {
   const { sourcifySource } = useAppConfigContext();
   const content = useContract(
@@ -46,7 +49,11 @@ const ContractFromRepo: React.FC<ContractFromRepoProps> = ({
         </div>
       )}
       {content !== undefined && (
-        <HighlightedSource source={content} langName={langName} />
+        <HighlightedSource
+          source={content}
+          langName={langName}
+          decorations={decorations}
+        />
       )}
     </>
   );
--- a/src/execution/address/Contracts.tsx
+++ b/src/execution/address/Contracts.tsx
@@ -3,6 +3,8 @@ import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { Menu, MenuButton, MenuItem, MenuItems } from "@headlessui/react";
 import { useQuery } from "@tanstack/react-query";
 import React, { lazy, useContext, useEffect, useState } from "react";
+import { useSearchParams } from "react-router";
+import { type DecorationOptions } from "shiki";
 import ContentFrame from "../../components/ContentFrame";
 import ExternalLink from "../../components/ExternalLink";
 import InfoRow from "../../components/InfoRow";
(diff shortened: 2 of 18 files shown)


sealer3:
New "revert trace button" style reveal:

image

When clicked:

image

Other addresses in the revert chain (including verified contracts without source mappings and unverified contracts) look like this:

image

sealer3:
I added chevrons to separate items in the revert trace and clarify the order of reverts:

image

wmitsuda:
@sealer3 could you update me on this one? I think it is now mergeable, right?

in that case it needs a rebase after all other merges.

sealer3:
We need to implement a SourcifyV2 backend before merging this one, and V2 is not out yet. (To test the functionality, I manually added extra outputs to the metadata.json of some local testing contracts.) The format will be slightly different when V2 is implemented, and the calls to the server will definitely need to change.

I think the review request then was for you to approve the design/style.

It looks like the "get verified contract" endpoint was just merged yesterday, so I might be able to start working on this soon: https://github.com/ethereum/sourcify/issues/1827#issuecomment-2623873770

wmitsuda:
ok

wmitsuda:
removing myself from review queue for now then :)

sealer3:
Note that because Solidity's source maps "choose" a single source location per instruction, if the optimizer is enabled and multiple code locations access the same instruction, it's possible that our mappings will be incorrect (because the compiler's source map is pointing to a different place that instruction is used).

I am able to prevent this from happening in the most common case by skipping over code blocks that from empirical observation are commonly deduplicated by the optimizer. (I used real on-chain transactions to find which ones were most common.) So, essentially all `revert`/`require` statements should work.

What could use additional testing is whether custom errors are deduplicated often and if they have a pattern we can detect if so. There weren't many contracts in my testing data that made use of them.

wmitsuda (approved):

PR #2583, opened 2024-10-30
Allow whatsabi to run when assetsURLPrefix is undefined

### LINKED ISSUE #2443: Allow whatsabi to run when `assetsURLPrefix` is undefined
The fix is relatively simple in `useWhatsabi.ts`. This would address #2441. We'll skip the 4bytes lookup in this case unless something like #2442 is configured.

sealer3:
Closes #2443


wmitsuda (approved):

wmitsuda:
it conflicted with another one I merged, rebase pls @sealer3

sealer3:
Done.

PR #2584, opened 2024-10-30
Recognize Solidity panic codes in transaction error messages

### LINKED ISSUE #2540: Handle Solidity panic codes
Failing assertions and other internal checks in Solidity contracts revert with `Panic(uint256)` and an opcode. We should display a human-readable revert message.

https://docs.soliditylang.org/en/latest/control-structures.html#panic-via-assert-and-error-via-require

sealer3:
Closes #2540.

Example:
image


wmitsuda:
suggestion: since those are Solidity reserved errors, what if we add a Solidity icon as a decoration, and let it explicit it is a Solidity panic exception in the error msg?

feel free to elaborate a good formatting.

wmitsuda:
also, since those codes are kind of rare/internal, what about we link it somewhere to the same Solidity docs you mentioned in the original issue? would it become too bloated?

sealer3:
These are great suggestions, and I think we can do them without much bloat.

sealer3:
@wmitsuda What do you think about this?

image

wmitsuda:
> @wmitsuda What do you think about this?
> 
> image

Still a little noisy, what about the following template:

"<error icon><solidity icon> Panic <code in hexadecimal>: <description> (docs)" -> link the word "docs" to solidity docs

the idea of both icons together is to be similar to the revert + blob icons from this block: https://holesky.otterscan.io/block/3052487/txs

sealer3:
This is the updated version:

image

wmitsuda:
@sealer3 do you mind pasting that tx hash here so I can see the page myself without having to type 64 characters? :)

wmitsuda:
ah, or is it some tx you made it yourself for some dev chain?

sealer3:
Yes, I made it myself. I can try to find a public mainnet example.

sealer3:
Here are some examples on mainnet:

- 0xa9532fa97afe1f031ab51b70bcb14877a00b29f4e306462adc66ee166e1cb1cd
- 0x6e901d091bbe9513908deb2293d2d45de38a07ce8297e0499a68b5b79e986e9d

It looks like overflow/underflows are the most popular panic on mainnet.

wmitsuda:
thanks, looks good!

PR #2666, opened 2024-11-29
Upgrade react-router to v7

sealer3:
The only required changes:
* `react-router-dom` becomes `react-router`
* `defer` is no longer needed (and no longer exists) since react-router automatically defers loading if you return a Promise in a data loader.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "otterscan",
   "version": "2.7.0",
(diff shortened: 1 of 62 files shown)


sealer3:
The e2e tests caught what looks like a relative Link which needs to be changed for v7... working on it...

wmitsuda (approved):

wmitsuda:
got conflicted because other merges I did, rebase pls?

PR #2712, opened 2024-12-25
Make aggregation bits grid columns responsive

sealer3:
Scales the aggregation bits on slot pages so they look better on small-width viewports:

image

### DIFF
--- a/src/components/ContentFrame.tsx
+++ b/src/components/ContentFrame.tsx
@@ -12,7 +12,7 @@ const ContentFrame: FC<PropsWithChildren<ContentFrameProps>> = ({
 }) => {
   return tabs ? (
     <div
-      className={`sm:divide-y rounded-b-lg border bg-white px-3 ${
+      className={`divide-y rounded-b-lg border bg-white px-3 ${
         isLoading && "opacity-50 transition-opacity"
       }`}
     >
--- a/src/components/ContentFrame.tsx
+++ b/src/components/ContentFrame.tsx
@@ -12,7 +12,7 @@ const ContentFrame: FC<PropsWithChildren<ContentFrameProps>> = ({
 }) => {
   return tabs ? (
     <div
-      className={`divide-y rounded-b-lg border bg-white px-3 ${
+      className={`sm:divide-y rounded-b-lg border bg-white px-3 ${
         isLoading && "opacity-50 transition-opacity"
       }`}
     >
--- a/src/components/InfoRow.tsx
+++ b/src/components/InfoRow.tsx
@@ -5,9 +5,9 @@ type InfoRowProps = React.PropsWithChildren<{
 }>;
 
 const InfoRow: React.FC<InfoRowProps> = ({ title, children }) => (
-  <div className="grid grid-cols-4 py-4 text-sm">
-    <div>{title}:</div>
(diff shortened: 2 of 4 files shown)


wmitsuda:
hm, I'm trying some presets on Chrome devtools, at least on "iPhone SE" one it still looks pretty scrambled.

I think it should be ok to let horizontal scroll on very narrow resolutions.

image

wmitsuda:
also, something I've been thinking of, what about in those very narrow resolutions we adopt some alternate layout, instead of:

<separator>
Section name:  <separator> <content>
<separator>
Section name:  <separator> <content>
<separator>
Section name:  <separator> <content>

it collapses into:

<separator>
Section name:
<content>
<separator>
Section name:
<content>
<separator>
Section name:
<content>

so we don't waste space below "Section name".

but that would be done page-wise, in all pages, but I think that can be done using a mix of tailwind classes so we don't need to bloat the code with grid conditionals.

wmitsuda:
(let me know if the idea above was not clear, otherwise I can draw something to better express what I'm thinking of)

sealer3:
I think I've got it. I also have manual scrolling for the aggregation bits if they overflow:
Screen Shot 2024-12-31 at 19 56 26

It affects all InfoRow components across the app:
Screen Shot 2024-12-31 at 20 04 37

This is for widths smaller than [`sm`](https://tailwindcss.com/docs/responsive-design), so for "regular" screen sizes we still have the grid layout.

wmitsuda:
I think it is better to reduce the number of columns in the aggregation bits instead of letting it overflow, i.e., in the screenshot it should be better to cap it to 4 columns.

wmitsuda:
another improvement: in this "stacked" layout, we should give some extra pixels (1 or 2?) vertical space between the section title and the content to give it some breath

wmitsuda:
and what about removing that vertical separator in the stacked layout? I think it only makes sense if everything is in one line, but when you stack the components, the line break between the stacked components fulfill that role

wmitsuda:
about the transaction hash, what about styling it with word break? I see no solution for such long strings containing hashes, but I feel forcing a word break at the limit is better than letting it overflow + horizontal scroll.

sealer3:
Transaction hash `break-all` example:
image

Small width:
image

sealer3:
This is the transactions view with the vertical space added and without dividers in the stacked view:



Personally, I prefer having the dividers there.

wmitsuda:
> This is the transactions view with the vertical space added and without dividers in the stacked view:
> 
> 
> 
> Personally, I prefer having the dividers there.

you can keep the dividers, I meant keep the dividers, but add space between, i.e., "Transaction Hash:" line and the hash itself.

because the content can have 1+ lines.

sealer3:
By the way, I'm working on this but having a hard time aligning it to the second (last) line. Aligning to the first line is easy:

image

Trying to put it at the end has it sinking below the baseline slightly:
image

wmitsuda:
hm, I think we can align it to the first line, it doesn't look so ugly. It's better than having it floating in the middle

sealer3:
Done.

wmitsuda (approved):
alright, let's merge this one. there are other individual fields still not responsible, but we can keep going with other individual PRs.

PR #2765, opened 2025-01-20
Slide in one-day price change from right with motion

### LINKED ISSUE #2691: Add some smooth animation to the price box pop-in
The price box at the top of the page is cool, but it is loaded async and the "pop-in" effect is really starting to distract me :)

I think we should make it smoother by adding some non-distracting animation when it is loaded.

let me try to attach a video...

sealer3:
Closes #2691

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,6 @@
         "@fortawesome/free-solid-svg-icons": "^6.5.2",
         "@fortawesome/react-fontawesome": "^0.2.2",
         "@headlessui/react": "^2.2.0",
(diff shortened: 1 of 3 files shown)


sealer3:
This change only slides in the one-day price change, which from the video takes longer to load. I can also have the price box load using an "appear" animation that brings it from a scale of 0 to 1 in a "pop into existence" format. I tried it out but thought the price change needed an animation the most. I can add that to this PR if you think it's required. I thought the slide in fit better with the price change than the native token price, but I can easily swap the two.

wmitsuda (approved):
looks great!

Issue #2740, opened 2025-01-06
Allow more Sourcify sources, add link to docs


sealer3:
Per https://github.com/otterscan/otterscan-book/issues/21, we should:

- [x] support multiple Sourcify sources (switcher should be used if more than one source is configured)
- [x] add an additional menu item linking to the docs explaining how to configure a local datasource
- [ ] explain the benefits of using local Sourcify sources (privacy, speed, data ownership)

PR #2793, opened 2025-01-25
Restructure Sourcify source configuration

sealer3:
Closes https://github.com/otterscan/otterscan/issues/2740

After we merge this, we'll need to update the docs to reflect the changes to the config. I made this version is backward-compatible with the previous config by checking if a string instead of a new SourcifySource object was passed into the config, but I added a deprecation warning and a link to the new docs.

Sources in the config object are ordered from top to bottom as specified in the config, with the first being the default. The key name is now the displayed name rather than an "internal" name.

If no sources are specified in the config, the default Sourcify Servers source is used, and it adds a link to the docs in the menu:
image

Let me know if you have other ideas for how this should look.

### DIFF
--- a/cypress/support/devnet-config.json
+++ b/cypress/support/devnet-config.json
@@ -4,8 +4,10 @@
   "assetsURLPrefix": "http://localhost:5175",
   "sourcify": {
     "sources": {
-      "central_server": "http://localhost:7077",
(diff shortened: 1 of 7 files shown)


wmitsuda (approved):

wmitsuda:
let's merge this one

PR #2794, opened 2025-01-26
Migrate to Tailwind v4

### LINKED ISSUE #2791: Migrate to Tailwind v4
We'll start using Tailwind v4.

- [x] Review differences between v3 and v4
- [x] Follow migration docs in case any of the breaking changes affect us

sealer3:
Closes #2791


sealer3:
The padding on `divide-dotted` seems to be gone in Tailwind v4 (todo):

image

sealer3:
The reason for the failing devnet test case was that the `.log` file in the Erigon devnet datadir kept updating, and this caused vite to issue page reloads whenever it noticed an update. It might be due to Tailwind's new source discovery system. I changed the location of the Erigon datadir, and it is no longer scanning that log file.

wmitsuda:
could you try to remove postcss as well and use tailwindcss as a 100% vite plugin?

reference: https://tailwindcss.com/docs/upgrade-guide#using-vite

I'm 99% it is safe, as the only reason postcss is a dependency is because tailwind used to require it, but needs test.

anyway, less dependencies == good, so we should try it.

sealer3:
I removed it from the package.json, but it appears to be a peer dependency.

wmitsuda:
I think the dependency comes from vite itself, but I think that's ok because I assume it won't be loaded and tree shaken.

wmitsuda:
see `npm ls -a`

PR #2808, opened 2025-01-28
Migrate to Ubuntu 24.04 for remaining actions

### LINKED ISSUE #2807: Migrate all actions workflows to use explicit ubuntu image versions
We should migrate all remaining github actions workflow files to use ubuntu-24.04 instead of ubuntu-latest and test them in the process.

Ubuntu 24.04 is now the default, but we should future proof ourselves from automatic rollouts.

BTW, apparently the automatic rollout broke the docker publishing task (https://github.com/otterscan/otterscan/pull/2806) so I have set it to ubuntu-22.04 for now in order to make it work again.

sealer3:
Closes #2807

### DIFF
--- a/.github/workflows/autofix.yaml
+++ b/.github/workflows/autofix.yaml
@@ -8,7 +8,7 @@ permissions:
 
 jobs:
   autofix:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


sealer3:
The pages-deployment action only runs when triggered manually. Everything else seems to be functional on ubuntu-24.04.

wmitsuda (approved):

Issue #1887, opened 2024-03-17
On really wide screens, tx search should give tx hash column more space


wmitsuda:
This is a little tricky, on really really wide displays, there is a lot of unused horizontal space, yet the tx hash column is currently truncated.

We should give it more horizontal space in this case so it can be displayed in full.

<img width="2553" alt="image" src="https://github.com/otterscan/otterscan/assets/28685/f65e36bf-6dba-4ff2-a657-bd4dac648209">

Issue #2796, opened 2025-01-27
Columns keep "jumping" in the transaction search results


wmitsuda:
- Search for vitalik.eth on mainnet
- Keep moving on through the pages
- Notice how the column sizes keep "jumping", and they are also irregular, I guess because they are adjusted by their contents.

See video:

https://github.com/user-attachments/assets/40ab8282-987d-4cfd-b12e-9cccfefcaed7

I didn't see this behavior when browsing through result pages from a block, only when searching by address.

Anyway, we should make the layout deterministic, so the main grid is predictable for user's eyes, then we adjust the content. If the content looks bad or doesn't fit well, then we should rework the main layout overall.

wmitsuda:
I think part of the blame is the "age" column, whose text gets bigger than the regular column size for some older dates.

wmitsuda:
I'd prefer consistency of column widths.

sealer3:
This is more or less fixed in #2815.

PR #2815, opened 2025-01-29
Set column sizes for tx hash/from/to in wider resolutions

wmitsuda:
fix https://github.com/otterscan/otterscan/issues/1887

tailwind 4 defines up to 2xl breakpoint, this PR defines 3xl and 4xl for the next 3 popular widths: 1920 and 2560 according to https://screensiz.es/

but we'll only use 4xl here, 3xl does not give enough free space to show everything.

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -34,6 +34,7 @@
 
   --breakpoint-3xl: 120rem;
   --breakpoint-4xl: 160rem;
+  --breakpoint-5xl: 200rem;
 }
 
 /*
--- a/src/index.css
+++ b/src/index.css
@@ -31,6 +31,9 @@
   --background-color-skin-from: rgb(var(--color-from-fill));
   --background-color-skin-to: rgb(var(--color-to-fill));
   --background-color-skin-table-hover: rgb(var(--color-table-row-hover));
+
+  --breakpoint-3xl: 120rem;
+  --breakpoint-4xl: 160rem;
 }
 
 /*
--- a/src/search/ResultHeader.tsx
+++ b/src/search/ResultHeader.tsx
@@ -16,8 +16,8 @@ const ResultHeader: FC<ResultHeaderProps> = ({
     <th>Method</th>
     <th className="w-28">Block</th>
     <th className="w-36">Age</th>
-    <th className="4xl:w-md">From</th>
-    <th className="4xl:w-md">To</th>
+    <th className="4xl:w-md 5xl:w-xl">From</th>
+    <th className="4xl:w-md 5xl:w-xl">To</th>
     <th className="min-w-52">Value</th>
     <th>
       <button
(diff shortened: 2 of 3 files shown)


wmitsuda:
Vitalik's txs on my mbp display (breakpoint NOT activated)

image

the same page on my external 2560 wide monitor.

image

sealer3 (commented):
There's a noticeable jump in the tx hash column width at the breakpoint. Please take a look at that.

Also, try testing this out on addresses with several tags. Here's an example of an address that now will no longer fit completely in the search results even on ultra-wide screens: `/address/0xB28d4B8653eeDA12Cb8B5fD202Bd18dADa156be1`

I don't have a strong preference regarding these examples, but I thought you should know.

wmitsuda:
> There's a noticeable jump in the tx hash column width at the breakpoint. Please take a look at that. (Update: Actually I guess this is the whole point of the PR...)
> 

that was intentional, it fits 1 extra icon (blob or failed tx) and I left some extra room to maybe 1 more.

> Also, try testing this out on addresses with several tags. Here's an example of an address that now will no longer fit completely in the search results even on ultra-wide screens: `/address/0xB28d4B8653eeDA12Cb8B5fD202Bd18dADa156be1`
> 

hm, right, that should not happen, lemme try some extra constraints.

wmitsuda:
ah wait, are you sure you tried it with the PR on?

image

sealer3:
Yes, but I have `experimental: true` in my Otterscan config. Here's what it looks like on an ultra-wide browser configuration:

image

wmitsuda:
ah, interestingly, mine is not showing the [165] tag, and the size fits exactly, so I don't see it truncating.

but does it respect the tx hash column size?

wmitsuda:
so, I think in the ultra-wide case it should have a larger size for the addresses, so the value is shrank.

what is the resolution of your monitor? mine is not ultra-wide, just a 5K monitor.

wmitsuda:
BTW, for easy detection of your dimensions regardless of retina display, zooming, etc..., better go to https://whatismyviewport.com/

sealer3:
Yes, the txn hash column is fine:
image

I don't have a large monitor, but I can make the size of my browser window larger than what is visible on my screen. The browser is reporting a viewport width of over 3200px for the size I used when I tested earlier. I wanted to make sure it wasn't just my screen width that was cutting the address off.

wmitsuda:
3200 seems to be the next standard resolution after 2560 according to: https://upload.wikimedia.org/wikipedia/commons/0/0c/Vector_Video_Standards8.svg

so let's create a new breakpoint at that width and for that we'll give more space to from/to columns.

image

sealer3 (approved):

sealer3:
I'm overloading this PR to adjust the from/to column widths :)

Please review the changes to that (addressing #2796).

wmitsuda:
looks ok

PR #2824, opened 2025-01-31
Remove top/bottom Whatsabi alert border

### LINKED ISSUE #2818: Error msg now have a strange bottom border
maybe some regression from tailwind 4 migration or something else. once identified the root cause we should check if other places were affected as well.

Image

sealer3:
Fixes #2818

### DIFF
--- a/src/execution/address/WhatsabiWarning.tsx
+++ b/src/execution/address/WhatsabiWarning.tsx
@@ -4,7 +4,7 @@ import React from "react";
 
 const WhatsabiWarning: React.FC = () => (
   <div
-    className="bg-orange-100 border-l-4 border-orange-500 text-orange-700 px-3 py-2.5 mb-4 rounded-sm mt-2"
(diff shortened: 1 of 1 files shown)


wmitsuda (approved):
let's merge this one, but I wonder if we should fix that on ContentFrame because it is inheriting this behavior from the parent and this fix sounds a little like a workaround.

I think when I first made ContentFrame I imagined it having content sections as children, hence the divider makes sense. But the warning div is something different.

sealer3:
You're right, ReadContract is a peculiar case of ContentFrame: it mostly has one large div item. Do you think it would make sense to add an option to ContentFrame that disables `divide-y`?

wmitsuda:
what I was thinking of is ContentFrame be turned into a more complex master component that supports an optional generic "warning messages" section that doesn't have divide-y, but yes the outcome would be like you said. wdyt?

PR #2827, opened 2025-02-03
Show USD conversion in Value column on wide screens

### LINKED ISSUE #2797: On wide resolutions, show the equivalent USD value in the "Value" column of search results
As can be seen in the following screenshot, there is plenty of horizontal space on wider resolutions (this is not even an ultra-wide monitor).

As we are already working on making ots good on small devices, we should on the other hand take advantage of higher resolutions to offer a better UX.

For this specific case, I think we could display the equivalent USD value next to the ETH value in the "Value" column, if:

- the column size allows for it (I think we could exploit the container query support of tailwind 4 to make it dynamic and appear only if the cell is big enough to show it, regardless of the display resolution).
- there is a chainlink oracle for the native token <-> USD (same criteria we use for that box at the top).

Challenges:

- Worst case scenario would have to calculate the USD value 25 times (let's say search results return 25 results, each one at a different block number, all of them != 0 ETH), UI may look slow and increase load on erigon, but let's try it. If we implement async load, it should not block UI, max make it progressively update slower.
- This feature will self-activate only on wide resolutions, even my macbook display doesn't have enough horizontal space to display it, but it should be desirable to make this info somehow available in smaller resolutions. I think that's ok for this issue, I have an idea for smaller resolutions already, will open another issue for that.

Image

sealer3:
Closes #2797.

At a client width of 1982px, the USD labels appear. There is the extra network overhead which is applied **at all resolutions** now, but the USD price of the native token is used elsewhere, so it has the side effect of preloading the "Txn Fee (USD)" column. (And you have the idea for smaller resolutions.)

image

### DIFF
--- a/src/search/TransactionItem.tsx
+++ b/src/search/TransactionItem.tsx
@@ -1,5 +1,6 @@
 import React, { useContext } from "react";
 import BlockLink from "../components/BlockLink";
+import { feePreset } from "../components/FiatValue";
 import MethodName from "../components/MethodName";
 import NativeTokenAmount from "../components/NativeTokenAmount";
 import NativeTokenAmountAndFiat from "../components/NativeTokenAmountAndFiat";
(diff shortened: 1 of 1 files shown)


wmitsuda (approved):
lgtm

PR #2893, opened 2025-02-13
Test that tx USD value is shown at wide viewports

### LINKED ISSUE #2852: Resolution-specific UI tests
This is a follow up of https://github.com/otterscan/otterscan/issues/2797

In that case, we now have a situation where a specific value (USD equivalent of ETH value) is displayed when resolution is wide, otherwise it is not, so I think we need to have some UI test to make sure this behavior is not broken.

Thinking long term, I imagine we'll have more and more of these UI scenarios where we'd like to provide better usage of screen space for users that have larger monitors and at the same time provide better UX on phones. Having UI tests for those are a requirement to keep our mental sanity :)

So, I'd like to propose using that previous issue as an exercise to:

- Research cypress docs and find a sane way to do resolution-specific tests (i.e., test code well organized, not spaguetti)
- Implement some UI test for the "regular" and "wide" scenarios of search results page (show/do not show the USD value along the ETH amount)

sealer3:
Closes #2852


wmitsuda:
ok, so the cy.viewport does the magic. I think it would improve the readability if we split the test in 2, wdyt?

so it becomes clear the intention of which test, smaller viewport does not have the box, bigger viewport has it, and you don't need to read the test logic to discover what each case does.

we can make it even clearer by using describing the intention/scenario in the test name, i.e. instead of "Search results", we can use "USD value must not be shown in smaller viewports"

sealer3:
You're right. I extracted deploying the price oracle into its own command and split the test into two. It looks better now.

wmitsuda (approved):

PR #2909, opened 2025-02-15
Make subtitle and tab header scrollable

### LINKED ISSUE #2861: Search results on iPhone 14 Pro Max viewport size
The most prominent issue here is the tab row, I think we could make it scrollable horizontally, and have some way to signal it is scrollable by having some fade effect on the left/right borders, not sure if have seen it already in other places, if you didn't, I can try to find an example.

Once we fix that, I think the address text will need some work as well, open to suggestions.

Image

sealer3:
Closes #2861 

This doesn't add the fade lines yet.


sealer3:
Yes, that will probably look better.

wmitsuda (approved):

PR #2910, opened 2025-02-16
Add USD toggle to tx Value column

### LINKED ISSUE #2853: Allow display of USD values on non-wide resolutions
This is an enhancement of https://github.com/otterscan/otterscan/issues/2797

This is a proposal for displaying the USD-equivalent values on search results on non-wide resolutions (due to the lack of horizontal space).

For non-wide resolutions, turn the "Value" column into a toggle similar to the "Txn fee", so the user can click on it to alternate between the "native token" value and the "USD value".

When in "USD value" mode, we can show it using the same "red-pill" like control we use on wide resolutions to keep consistency instead of a plain text.

sealer3:
This makes the Value column header clickable into a toggle like the Txn Fee column:
image

When clicked:
image

When the space is large enough that both fit, the header is no longer clickable:
image

I added reduced-opacity $0.00 entries when no value was sent and made a more generic `useToggler`, though it relies on the TypeScript compiler's implementation of enums (due to https://github.com/microsoft/TypeScript/issues/42457 - it's at least better than other attempts like https://www.npmjs.com/package/enum-for?activeTab=code in my opinion).

Closes #2853

### DIFF
--- a/src/components/NativeTokenAmountOnlyFiat.tsx
+++ b/src/components/NativeTokenAmountOnlyFiat.tsx
@@ -1,0 +1,38 @@
+import { BlockTag, FixedNumber } from "ethers";
+import { FC } from "react";
+import { useFiatValue } from "../usePriceOracle";
+import FiatValue, { FiatBoxProps, neutralPreset } from "./FiatValue";
+
+type NativeTokenAmountOnlyFiatProps = FiatBoxProps & {
+  value: bigint;
+  blockTag?: BlockTag;
+};
+
+/**
+ * A variant of NativeTokenAmountAndFiat that does not show the native token amount.
+ */
+const NativeTokenAmountOnlyFiat: FC<NativeTokenAmountOnlyFiatProps> = ({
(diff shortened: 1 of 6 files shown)


wmitsuda:
should we make the mouse pointer change to "hand" on hover? if so, we should do the same for the "txn fee" column.

sealer3:
> should we make the mouse pointer change to "hand" on hover? if so, we should do the same for the "txn fee" column.

You're right, this comes from Tailwind v4 using the browser default for the cursor: https://github.com/tailwindlabs/tailwindcss/issues/8961

I think the argument in that issue is sound, that the CSS spec says that `pointer: cursor` indicates a link: https://drafts.csswg.org/css-ui-3/#valdef-cursor-pointer

We already change the text color on hover, which I think is enough of an affordance that the table header is clickable. So, in my opinion, we should just leave it as-is.

wmitsuda (approved):
ok, let's merge it

PR #2959, opened 2025-02-26
Load 4bytes in whatsabi feature when assetsURLPrefix is empty

### LINKED ISSUE #2957: whatsabi not loading 4bytes if `assetsURLPrefix` is an empty string
The default Docker configuration sets `assetsURLPrefix` to the empty string. The whatsabi loader should not skip the 4bytes lookup if `assetsURLPrefix` is empty (only do that if it's null or undefined).

Relevant code:
https://github.com/otterscan/otterscan/blob/925fd74f5dadba3eb84b49b2d0faa9018a3ec183/src/sourcify/useWhatsabi.ts#L40

sealer3:
Closes #2957 

Still needs testing on the official Otterscan docker version.

### DIFF
--- a/src/sourcify/useWhatsabi.ts
+++ b/src/sourcify/useWhatsabi.ts
@@ -37,7 +37,7 @@ function whatsabiFetcher(
     const selectors = whatsabi.selectorsFromBytecode(code);
     const decodedFunctions: (string | null)[] = await Promise.all(
       selectors.map(async (selector) => {
-        if (!assetsURLPrefix) {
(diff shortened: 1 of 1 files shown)


wmitsuda:
FYI: you can generate the docker image locally if you need

sealer3:
Yes. Building the image locally, I can confirm this is able to show whatsabi signatures using the 4bytes database with just the erigon URL set in the environment variables.

wmitsuda (approved):

PR #3000, opened 2025-03-08
Add initial SourcifyV2 support and use as default

sealer3:
Assumes RepositoryV1 is also enabled on the server, which is the only way to fetch sources currently.

This also fixes what appears to be a bug if fetching metadata when the app's sourcifySource is null.


sealer3:
This PR is integrated into #2579.

PR #3080, opened 2025-04-01
Overhaul address search

### LINKED ISSUE #3036: Address page misses transactions if more than page size are received in a single block
See https://hoodi.otterscan.io/address/0x889878C72f56C39F7772e417B489c1A56F76D45c

This address received 100 transactions in one block, but the address page only shows 25 of them.

sealer3:
Closes #3036.

Overhauls the address search to handle pages with overflow correctly. This PR fixes multiple bugs:
* Navigating prev/next on overflow pages did not work correctly (transactions were skipped)
* First/last page markers when there is overflow (ex: if the first query has > `PAGE_SIZE` transactions but was both `firstPage` and `lastPage`, the UI would not show more than the first `PAGE_SIZE` transactions)
* Intra-block transaction search URL (shown transactions could differ when navigating to that URL if there was more than one transaction from the same block)

The key insight is that queries return all transactions present in each different block number in the response, so we preserve these queries as "batches" which lets us determine if we're on the first or last page even when paging between batches.

### DIFF
--- a/cypress/e2e/mainnet/address-search.cy.ts
+++ b/cypress/e2e/mainnet/address-search.cy.ts
@@ -1,6 +1,7 @@
 describe("Advanced search", () => {
   const navigateAndAssert = (
     initialUrl: string,
+    initialHash: string,
     navAssertions: [direction: "next" | "prev", expectedHash: string][],
     txHashIndex: number,
   ) => {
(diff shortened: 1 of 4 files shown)


sealer3:
I added test cases, which I've confirmed by manually making ots RPC calls and stepping through the transactions to ensure correctness. I think it's ready for review.

wmitsuda (approved):

PR #3153, opened 2025-05-03
Convert to React Router framework mode

sealer3:
This PR migrates Otterscan to React Router's [framework mode](https://reactrouter.com/start/modes). Server-side rendering is currently set to `false` for SPA mode, but as we migrate fetching into loaders, this will change. (It seems like it even works if you set `ssr: true` today in `react-router-config.ts`.)

Related issue: #2388 

Required follow-up work: migrating the loaders previously found in App.tsx, which are now removed. It's possible that since all these loaders are currently just prefetching state and require access to the provider, we won't implement them as `clientLoader`s but instead wait until we migrate them to become `loader` functions (server-side).

Notes:
* `useKeyboardShortcut` does not seem to be compatible with `react-router` due to library-level import, so I've commented out its use. We could potentially bring it into our code base since it's small (150+88 lines of code).
* The build folder and scripts are not the same! We run with `react-router` instead of vite. The build folder is `./build/client` for client scripts and `./build/server` if `ssr: true` for server scripts.
* React has been updated to React 19 for React Router support. `react-helmet-async` hasn't been updated in over a year and needed a dependency override in `package.json`.

Let's make sure we test the build process because of this...


sealer3:
After a lot of testing and difficulty reproducing the error, it seems like it's due to vite not detecting the dependencies properly and then forcing a reload in the middle of the test:

```
4:49:02 PM [vite] (client) ✨ new dependencies optimized: react-content-loader, @shazow/whatsabi, react-blockies
4:49:02 PM [vite] (client) ✨ optimized dependencies changed. reloading
    1) Should load vitalik.eth address
```

PR #3311, opened 2025-06-20
Highlight entire line of revert locations

### LINKED ISSUE #3293: Highlight the entire line on revert location
A follow up of https://github.com/otterscan/otterscan/pull/2579

Would it be possible to highlight the entire line? We can keep the current highlight of the revert reason, but I find it hard for my eyes to find it among all noise.

Highlighting the entire line in a different color (in addition to the revert hightlight) would help.

Example:

Image

I think a light red background would make it easier to guide user's eyes to the correct place.

If it is not possible/shiki doesn't support it easily/too hard, we can close this one.

sealer3:
Closes #3293 

Makes the highlighted region more visible in dark mode and adds a red background for the rest of the line(s).

Image

Image

### DIFF
--- a/src/execution/address/Contracts.tsx
+++ b/src/execution/address/Contracts.tsx
@@ -65,6 +65,7 @@ const Contracts: React.FC<ContractsProps> = ({ checksummedAddress, match }) => {
     start: number;
     end: number;
   } | null>(null);
+  const [highlightLines, setHighlightLines] = useState<number[] | null>(null);
   useEffect(() => {
     const hrParam = searchParams.get("hr");
     if (hrParam) {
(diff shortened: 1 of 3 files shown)


wmitsuda (approved):

sealer3:
@wmitsuda I put in a request for a simple feature in shiki that would make this easier to implement: https://github.com/shikijs/shiki/issues/1038

Once that's merged, then we just need to add `w-full inline-block` to get the desired effect:
image

wmitsuda:
great! thanks!

sealer3:
Unfortunately it doesn't look as good with multiple lines:
image

The gaps are from highlighted green portions that are not `inline-block`. I suppose we could make those `inline-block` and get this:

image

Single line:
image

What do you think?

wmitsuda:
I think it looks ok-ish, we can do this way if it doesn't break other things

PR #3315, opened 2025-06-21
Jump straight to block number in search results

### LINKED ISSUE #930: Enable address transaction page navigation by block number in addition to tx hash
I was browsing the transaction history of an address that made over 100,000 transactions. I wanted to navigate to a page that contained a specific block number (say 12345678).

When you use the controls, `UndefinedPageControl`/`UndefinedPageButton` exposes a transaction hash in the query parameters (`/address/${address}/txs/${direction}?h=${hash}`), so to be able to navigate to the page with block 12345678, I navigated to `/block/12345678`, clicked the transactions button, copied the first transaction hash in the block, and finally pasted it into the query parameter for `h`. Finally, I was on the page I wanted, even though this hash was not sent by the right address!

So I propose we allow some other query parameter like `?b=` or `?block=` that fetches the first transaction in the block and redirects to the `?h=${hash}` page. That way, it's easy to navigate to the right page by block number and not just transaction hash.

### LINKED ISSUE #3316: Address search results sometimes do not update with browser navigation
Even though we recently updated the address search, it looks like this bug has been around for a very long time (as far back as is compatible with Erigon v3).

To reproduce:
1. Go to an address page that has at least 3 pages of address results.
2. Hit the "next" button 2 times.
3. Use your browser's back button.

The address search results do not update to reflect the URL change.

It seems like the reason is that the SearchController's `nextPage` and `prevPage` functions are responsible for stopping the navigation, and if the hash in question is not inside the current page of transactions, the function does nothing. What complicates the matter is that the `AddressTransactionResults` runs its update on a `useEffect` that depends on the controller object, so if `nextPage`/`prevPage` returns a new object, the `useEffect` gets run again. This does not affect the first and last pages because we can easily check if it's on the first or last page: https://github.com/otterscan/otterscan/blob/5470f8598c5bee30d60bef62e8a273c078351086/src/execution/address/AddressTransactionResults.tsx#L109-L128

sealer3:
Accepts a `?b=<block number>` query parameter for "middle" address search results, which adjusts the search controller to search starting from that block. If the block is too high/low, the first/last page is returned.

Based on top of #3317 

Closes #930 
Closes #3316

### DIFF
--- a/cypress/e2e/mainnet/address-search.cy.ts
+++ b/cypress/e2e/mainnet/address-search.cy.ts
@@ -1,21 +1,27 @@
 describe("Advanced search", () => {
   const navigateAndAssert = (
+    address: string,
     initialUrl: string,
     initialHash: string,
-    navAssertions: [direction: "next" | "prev", expectedHash: string][],
+    navAssertions: [
+      direction: "next" | "prev" | "back",
+      expectedHash: string,
+    ][],
     txHashIndex: number,
   ) => {
     cy.visit(initialUrl);
-    cy.get('[data-test="address"]', { timeout: 15_000 }).contains(
-      "0xdAC17F958D2ee523a2206206994597C13D831ec7",
-    );
+    cy.get('[data-test="address"]', { timeout: 15_000 }).contains(address);
     cy.get('[data-test="tx-hash"]')
       .eq(txHashIndex)
       .invoke("text")
       .should("equal", initialHash);
 
     navAssertions.forEach(([direction, expectedHash]) => {
-      cy.get(`[data-test="nav-${direction}"]`).first().click();
+      if (direction === "back") {
+        cy.go("back");
+      } else {
+        cy.get(`[data-test="nav-${direction}"]`).first().click();
+      }
       cy.get('[data-test="tx-hash"]')
         .eq(txHashIndex)
         .invoke("text")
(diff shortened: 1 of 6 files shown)


wmitsuda:
I'd remove the up/down buttons, I don't see much value in those since they increment units of 1, but you need to type the block number to begin with. I think they just add noise.

re: functionality, I have mixed feelings, at first sight I found it useful, but the UX feels counter intuitive, i.e., I type 15000000 and it brings a page with txs <= block 15000000. I feel like the what one would expect is to bring the page with txs >= 15000000.

sealer3:
> I'd remove the up/down buttons

Sure, I'll do this.

> bring the page with txs >= 15000000

Fortunately, this is an extremely simple change (`next` -> `prev`), so I can add this soon.

sealer3:
One thing to note is that if we show the transactions >= the target block number, we will be displaying the transactions closest to that block number at the *bottom* of the page rather than at the top. This is simply due to the way transactions are displayed (most recent to least recent).

wmitsuda:
yeah, what do you think? in a first thought I feel like this is less confusing than the current behavior

sealer3:
I personally prefer the current behavior. My reasoning is that when you perform a query, you usually expect the top results to be closest to your query. If you have a specific block number in mind, I think you should be able to see those results immediately.

This one is tricky because there are few web2 occurrences of this we can look to for guidance. It might be worth running a poll to see whether people would rather see the target block txs + earlier txs or the later txs + target block txs.

wmitsuda (approved):
alright, on a second thought I think your argument makes sense, lets merge it the current way.

PR #3317, opened 2025-06-23
Fix address search browser pagination

### LINKED ISSUE #3316: Address search results sometimes do not update with browser navigation

sealer3:
Closes #3316 and adds a mainnet test (which fails without this PR).

The fix involves checking additional parameters which are saved in the SearchController object.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -28,7 +28,7 @@ import { PAGE_SIZE } from "./params";
 import ProbeErrorHandler from "./ProbeErrorHandler";
 import { queryClient } from "./queryClient";
 import { loader as searchLoader } from "./Search";
-import { getTransactionQuery, searchTransactionsQuery } from "./search/search";
+import { searchTransactionsQuery } from "./search/search";
 import { SourcifySourceName } from "./sourcify/useSourcify";
 import { ConnectionStatus } from "./types";
 import { AppConfig, AppConfigContext } from "./useAppConfig";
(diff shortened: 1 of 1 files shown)


sealer3:
Ready for review. I noticed the prefetching in the loader function for middle pages was getting out of sync with the optimized search loader, causing lots of extra search queries that weren't being used, so that is now removed. It was giving us little benefit anyway: it prefetched one search query (of several required) if you directly navigate to a middle transaction page via a link.

PR #3322, opened 2025-06-23
Display a "code" icon for type 4 txs

wmitsuda:
Similar to blob txs, let's make it easier to see set code txs in the tx list


wmitsuda:
this is how it looks like:

image

sealer3 (approved):
Looks good.

PR #3356, opened 2025-07-01
Disable CCIP off-chain reads on ENS queries by default

### LINKED ISSUE #3330: Disable CCIP reads by default, even for ENS lookups
By default, ethers.js allows for "off-chain" [ERC-3668 CCIP reads](https://eips.ethereum.org/EIPS/eip-3668) during ENS resolveName requests: https://github.com/ethers-io/ethers.js/issues/2478

For privacy, we should disable these by default. We don't need users leaking information directly to Coinbase or other third parties for ENS lookups. We can add a setting in the configuration for users who want to have this functionality.

sealer3:
Closes #3330

I could not find the off-chain reverse lookup that was triggering CCIP reads during regular browsing, but I remember seeing these requests before. To test, try going to `nick.base.eth` or `jesse.cb.id` on the develop branch and see in the networking dev tools tab that it makes queries to Coinbase. Then, try the same in this PR. The name resolution should fail and no external lookups are attempted.

You can restore the CCIP read functionality (for ENS lookups only) in the config with the `enableOffchainEnsLookups` key, though I wonder if there's a better key name or section we can use, like "privacy".

### DIFF
--- a/src/useConfig.ts
+++ b/src/useConfig.ts
@@ -208,6 +208,13 @@ export type OtterscanConfig = {
   };
 
   /**
+   * Enable off-chain CCIP reads (EIP-3368) to third parties in smart contracts
+   * that support them, during ENS lookups only. It leaks your request to the
+   * URL specified in the CCIP read, so it is disabled by default for privacy.
+   */
+  enableOffchainEnsLookups?: boolean;
+
+  /**
    * Temporary config option, until address labels are complete: Enables setting
    * address labels which are kept in local storage.
    */
--- a/src/useConfig.ts
+++ b/src/useConfig.ts
@@ -207,12 +207,19 @@ export type OtterscanConfig = {
     l1ExplorerURL?: string;
   };
 
-  /**
-   * Enable off-chain CCIP reads (EIP-3368) to third parties in smart contracts
-   * that support them, during ENS lookups only. It leaks your request to the
-   * URL specified in the CCIP read, so it is disabled by default for privacy.
-   */
-  enableOffchainEnsLookups?: boolean;
+  externalDataSources?: {
+    ccip?: {
+      /**
+       * Enable off-chain CCIP reads (EIP-3368) to third parties in smart
+       * contracts that support them, during ENS lookups only. It leaks your
+       * request to the URL specified in the CCIP read. This should be disabled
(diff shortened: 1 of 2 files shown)


wmitsuda:
while I agree with this PR, I see a minor UX issue here, bc from now on if you try to resolve e.g. "nick.base.eth" it says the name does not exist, but in reality it should be better if we signal the user that we blocked the external URL resolution (+add some button to allow it or not, +add some link to docs on explaining how to enable all by default in the configuration).

is it possible to detect such thing on our interactions with ethers or is it a black box?

sealer3:
By default, it's a black box and returns `null` just as would be returned if the resolver didn't return anything: https://github.com/ethers-io/ethers.js/blob/9fd9d41d017a5e3b329aca47c79786e69cd40b99/src.ts/providers/abstract-provider.ts#L1239-L1240

According to this blog post, we could technically subclass `JsonRpcProvider` ourselves and create a custom error message when it's attempted: https://blog.ricmoo.com/highlights-ethers-js-march-2022-f511fe1e88a1

wmitsuda:
I think we should find a way to do it, either by "hacking" it ourselves or by asking ethers to provide a way to inform the caller if resolving the name should trigger an external read, so the caller can take a conservative approach and block the caller UI from triggering such call, or ask the user if he wants to proceed and the code should call the resolving method with some flag to resolve everything anyway.

wmitsuda:
because if we simply block everything that would be an UX issue since we are blocking popular domains like the base ones; and the user would think this is an error from Otterscan.

I think a good balance between privacy by default and allowing CCIP reads would be taking an approach similar to when gmail blocks images by default, but alert and allow the user to proceed. I think our UI could be similar.

image

sealer3:
While we wait for ethers support/response, I've enabled the setting by default as suggested. I think reverse-ENS lookups which trigger CCIP reads (not caused by user input) should be disabled by default or at least hidden behind a gmail-like notification in the UI. However, we are not yet able to configure that with ethers.

wmitsuda (approved):
good!

wmitsuda:
@sealer3 we need to add this to the docs as well

wmitsuda:
I agree that if we are able to do the gmail-like approval, then we should move it to be disabled by default.

PR #3407, opened 2025-07-23
Feature/kleros address tags display

gmkung:


wmitsuda:
@gmkung I found out this specific address looks weird: 0x40A50cf069e992AA4536211B23F286eF88752187

not sure if some data from it is making it appear empty on Otterscan, or the curated data is somehow wrong on Kleros side.

see:

image

wmitsuda:
Another thing I'd like to keep is having the Sourcify icon prepended to the control when the contract is verified by Sourcify, e.g., in the USDC contract; the contract is both verified by Sourcify and curated by Kleros.

I think in those cases we can keep Kleros icon/text style, but still prepend the Sourcify icon in order to give the user a quick insight that the contract is also verified by Sourcify.

@sealer3 something we need to improve after this PR is how to easily allow the user to switch between Sourcify/Kleros/ENS/etc. labelling. I have an idea, can show you later. For now just having some hints by keeping the icon/link next to the Kleros labelling works.

image

wmitsuda:
re: Kleros section in the transactions page, I think we should move it to a dedicated tab, WDYT @sealer3 ?

reasons for that:

- in order to not keep growing the "address overview" header (and pushing the search results down)
- Sourcify info is already in their own tabs
- I'm in doubt about what the tab should be called, though. For Sourcify stuff we don't call it "Sourcify", because we want to focus in the functionality first, but keeping the credits/branding of the data provider, i.e. the Sourcify tab is called "Contract", but with the Sourcify icon.
  - We should call the tab anything but Kleros, but keep the Kleros icon if we follow the same policy.
  - Don't know exactly what is the best name, I was thinking about "Description", "Project", "About", "Attestations", @gmkung do you have an opinion about what best describes those data that Kleros provides?

@gmkung BTW, for the "View on Kleros Scout" link and the dapp URL, you can use the "ExternalLink" component; it will apply the standard styling for external links.

image

wmitsuda:
BTW, another reason to move it to a different tab: there is more vertical space to unpack the context, i.e., it would be good to use that space to list all (dynamic) address-related dapp domains instead of putting them inside the tooltip.

wmitsuda:
I should have more suggestions still, but let's tackle those ones first. anyway, great work so far @gmkung !

sealer3:
@wmitsuda If there is Kleros information, I think it would be best to keep most of it on the address page, since it is important, and once ots2 is complete again, we'll already have enough ots2 tabs. That said, there's a lot we can do to compress the vertical space: as you mentioned, "View on Kleros Scout" can be turned into a vertical link or at least be placed on one of the previous lines, and token information can also be moved or removed.

In general the Copilot suggestions are accurate.

Also, before this is merged:
- [ ] Remove AI-generated `kleros-test.md`
- [ ] Remove changes to `public/config.json`
- [ ] `supportedChains` config option is unused
- [ ] Move `kleros` config section underneath a new `dataSources` (or similar) root config key
- [ ] Move inline SVG files into external files with names

wmitsuda:
@sealer3 have a look at the contract 0x1111111254EEB25477B68fb85Ed929f73A960582

it has 5 verified associated domains, it is currently shown as a tooltip, but I'd like not to use tooltips because they defeat the purpose of links which is to be clicked :)

but then we get the issue that it is a list with dynamic size, so let's say there are 10 domains, it'll push down the content a lot.

wmitsuda:
+ the text contents describing the project can be large, potentially shifting everything else down a lot (I'd like to avoid scrollbars here, I think it may make the UI noisy)

gmkung:
@wmitsuda thanks for looking at the PR, let us know if we can be of any help!

wmitsuda:
> @wmitsuda thanks for looking at the PR, let us know if we can be of any help!

I think the points we mentioned above should be addressed before we merge it

sealer3:
Regarding the unbounded list of verified domains, maybe they could be shown under a collapse/reveal button similar to the "extra call options" button in the Read Contract interface.

wmitsuda (changes requested):
marking PR as "changes requested"

gmkung:
Looking at the changes now, for the revert of public/config.json, should I move the 'kleros' object to something like this ? 

BEFORE:
```
{
  "erigonURL": "https://docs-demo.quiknode.pro",
  "assetsURLPrefix": "http://localhost:5175",
  "kleros": {
    "enabled": true,
    "apiUrl": "https://scout-api.kleros.link"
  }
}
```

AFTER:
```
{
    "erigonURL": "https://docs-demo.quiknode.pro",
    "assetsURLPrefix": "http://localhost:5175",
    "datasources":
    {
        "kleros":
        {
            "enabled": true,
            "apiUrl": "https://scout-api.kleros.link"
        }
    }
}
```

sealer3:
@gmkung Yes, that's right. Actually, could you rename it to `externalDataSources`? That would be even better. One of the emphases of Otterscan's readme page had been that data comes "all from your own node", so we should make the distinction clear.

gmkung:
@sealer3 

Ok the following changes were made. 

Config & Structure
✅ Renamed datasources → externalDataSources
✅ Removed unused supportedChains option
✅ Deleted kleros-test.md file
✅ Renamed 'Project' to 'Verified Info'
External Links
✅ Replaced inline <a> tags with ExternalLink component
✅ Consistent styling + external link icons
✅ Proper accessibility attributes
Verified Domains
✅ Replaced tooltip with collapsible LabeledSwitch
✅ Each domain now clickable as ExternalLink

Let me know if I missed anything.

public/config.json
```diff
@@ -1,4 +1,12 @@
 {
-  "erigonURL": "http://localhost:8545",
-  "assetsURLPrefix": "http://localhost:5175"
+  "erigonURL": "https://docs-demo.quiknode.pro",
```
wmitsuda: please, do not change the pre-configured erigonURL field, also apply the standard json formatter

wmitsuda:
I think Kleros should come enabled by default and pointing to their own operated server like we have Sourcify enabled by default and using their own hosted service, but allowing the user to disabled it or change the URL via configuration.

do you agree @sealer3 ?

wmitsuda:
Some suggestions for this section in order to make better use of screen space:

image

- I think you can move the "View on Kleros Scout" link to the "Verified Info" label; also can the link only "Kleros Scout" and move the icon to be next to the "Kleros Scout" label.
  - i.e.: from "-icon- Verified Info:" to: "Verified Info (-icon- Kleros Scout):"
- Remove the toggle + Show verified domains; always show the first 2 domains, if there are more than 2, replace the second line with a link "show N more domains", which when clicked replaces the link the the remaining list.
  - this way we ensure by default it'll take max 2 lines of info
  - for the address 0x1111111254EEB25477B68fb85Ed929f73A960582 it'll show:

```
- app.1inch.io
(show 4 more domains)
```
 
and when the user clicks the "(show 4 more domains)" link, it'll expand to the full list:

```
- app.1inch.io
- swapr.eth.link
- app.zerion.io
- swapr.eth.limo
- dex.guru
```

wmitsuda:
@gmkung what does the "Public name tag" mean?

In case of this 1inch contract it is "Router", but that alone doesn't sound much informative. in the example above it appears in the UI together with the contract name, which makes it confusing. I wonder if we either move it, improve presentation or don't show it at all.

image

wmitsuda:
also which of those info are optional?

wmitsuda:
answering my own question, I think all of them are optional, I think I found one such case: 0x881D40237659C251811CEC9c364ef91dC08D300C

@gmkung would it be a bug in Kleros service side or such all empty data is possible?

image

sealer3:
> I think Kleros should come enabled by default and pointing to their own operated server like we have Sourcify enabled by default and using their own hosted service, but allowing the user to disabled it or change the URL via configuration.
> 
> do you agree @sealer3 ?

I think this really depends on user expectations and what they want out of Otterscan. As we integrate more external services, we'll want to build a configuration wizard to help match users' goals, and then defaults will eventually matter less. If most of our users are running Otterscan for a local chain in Anvil or expect Otterscan to be private by default, the Kleros integration should be disabled. For users who are browsing Ethereum and want maximum insights, we should enable it.

So, for now I have a neutral opinion!

gmkung:
@wmitsuda Ok I've implemented the changes such that the list of Verified Domains will not exceed 2 lines! Also fixed the weird orphaned dot when there is no info returned besides verified domains

Regarding the optionality of data, indeed, not all fields are returned all the time, depending on what information is already provided by our community. That being said '0x881D40237659C251811CEC9c364ef91dC08D300C' should return more data, so let me troubleshoot the issue in our backend, but it shouldn't affect the interface of the current implementation.

wmitsuda:
> > I think Kleros should come enabled by default and pointing to their own operated server like we have Sourcify enabled by default and using their own hosted service, but allowing the user to disabled it or change the URL via configuration.
> > do you agree @sealer3 ?
> 
> I think this really depends on user expectations and what they want out of Otterscan. As we integrate more external services, we'll want to build a configuration wizard to help match users' goals, and then defaults will eventually matter less. If most of our users are running Otterscan for a local chain in Anvil or expect Otterscan to be private by default, the Kleros integration should be disabled. For users who are browsing Ethereum and want maximum insights, we should enable it.
> 
> So, for now I have a neutral opinion!

then let's make it enabled by default like Sourcify, in order to promote discoverability. but as in Sourcify, we offer and document ways to disable it.

wmitsuda:
> @wmitsuda Ok I've implemented the changes such that the list of Verified Domains will not exceed 2 lines! Also fixed the weird orphaned dot when there is no info returned besides verified domains
> 
> Regarding the optionality of data, indeed, not all fields are returned all the time, depending on what information is already provided by our community. That being said '0x881D40237659C251811CEC9c364ef91dC08D300C' should return more data, so let me troubleshoot the issue in our backend, but it shouldn't affect the interface of the current implementation.

then I think we need to handle the "all empty" case regardless. since the label is the "public name" field, can you make it act as if there is no kleros info in this case? this way it should fallback to other renderers, for this specific address it should automatically fallback to render Sourcify labelling.

wmitsuda:
> @wmitsuda Ok I've implemented the changes such that the list of Verified Domains will not exceed 2 lines! Also fixed the weird orphaned dot when there is no info returned besides verified domains

good, it is better now. the fields above still look a little cluttered, let me think about another UI solution for that.

image

src/kleros/KlerosAddressInfo.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+export default KlerosAddressInfo;
+
+// Always show first 2 domains, expandable for more
+const VerifiedDomains: React.FC<{ domains: string[] }> = ({ domains }) => {
+  const [expanded, setExpanded] = useState<boolean>(false);
+
+  if (domains.length <= 2) {
+    return (
+      <div>
```
wmitsuda: this code block and the next one are the same, no? we can remove the code duplication by collapsing the if

src/kleros/KlerosAddressInfo.tsx
```diff
@@ -0,0 +1,175 @@
...
+        <div className="text-sm text-gray-600 dark:text-gray-400 mb-1">
+          Verified domains:
+        </div>
+        <ul className="text-sm text-gray-700 dark:text-gray-300 list-disc pl-5 space-y-0.5">
```
wmitsuda: I think space-y-0.5 makes the UI too busy, let's change it to space-y-1 which looks more breathable from what I tested here

gmkung:
@wmitsuda ok, so if there is not tag info (but there are still verified domains) you want to ignore the verified domains and have it behave like there are no Kleros information at all?

src/kleros/KlerosAddressInfo.tsx
```diff
@@ -0,0 +1,175 @@
...
+    );
+  }
+
+  // Show first domain + "show N more" link
+  return (
+    <div className="text-sm text-gray-700 dark:text-gray-300">
+      <div className="text-sm text-gray-600 dark:text-gray-400 mb-1">
+        Verified domains:
+      </div>
+      <ul className="list-disc pl-5 space-y-0.5">
```
wmitsuda: same as the previous comment, let's use space-y-1 instead

wmitsuda:
> @wmitsuda ok, so if there is not tag info (but there are still verified domains) you want to ignore the verified domains and have it behave like there are no Kleros information at all?

yes, because when we are only showing/translating address -> labels, we are using only the kleros name; so if there is no kleros name, then there is nothing to show there (but if the user clicks the address, then other available kleros info like verified domains, etc, will be shown in their proper UI sections)

gmkung:
> > @wmitsuda ok, so if there is not tag info (but there are still verified domains) you want to ignore the verified domains and have it behave like there are no Kleros information at all?
> 
> yes, because when we are only showing/translating address -> labels, we are using only the kleros name; so if there is no kleros name, then there is nothing to show there (but if the user clicks the address, then other available kleros info like verified domains, etc, will be shown in their proper UI sections)

Oh do you want to move the entire KlerosAddressInfo into a different tab/popup? Maybe we could have a quick call and you can show me what you mean.

wmitsuda:
> > > @wmitsuda ok, so if there is not tag info (but there are still verified domains) you want to ignore the verified domains and have it behave like there are no Kleros information at all?
> > 
> > 
> > yes, because when we are only showing/translating address -> labels, we are using only the kleros name; so if there is no kleros name, then there is nothing to show there (but if the user clicks the address, then other available kleros info like verified domains, etc, will be shown in their proper UI sections)
> 
> Oh do you want to move the entire KlerosAddressInfo into a different tab/popup? Maybe we could have a quick call and you can show me what you mean.

no, I mean, case like this one below should not happen, if the UI code detects the Kleros label renderer would result in an empty string, the renderer should not be activated.

image

gmkung:
@wmitsuda Ok understood, pushed a fix!

wmitsuda:
@gmkung thanks, could you also rebase or merge `develop` tag to this one in order to certify this PR is working in sync with recent changes?

also could you please take a look at the copilot comments and resolve/close if they make sense.

wmitsuda:
> @wmitsuda Ok I've implemented the changes such that the list of Verified Domains will not exceed 2 lines! Also fixed the weird orphaned dot when there is no info returned besides verified domains
> 
> Regarding the optionality of data, indeed, not all fields are returned all the time, depending on what information is already provided by our community. That being said '0x881D40237659C251811CEC9c364ef91dC08D300C' should return more data, so let me troubleshoot the issue in our backend, but it shouldn't affect the interface of the current implementation.

I see `0x881D40237659C251811CEC9c364ef91dC08D300C` is now returning data, did you figure what it was?

gmkung:
@wmitsuda 0x881D40237659C251811CEC9c364ef91dC08D300C is returning data now due to an issue with the API endpoint in the background (incorrect handling of checksummed addresses). I discovered and fixed that now, so the problem for that address is fixed. But the situation where there are verified domains but no project labels is still a possible one, and the fixes should have addressed this now. 

Let me rebase and incorporate the relevant Copilot changes and get back to you with a final version.

gmkung:
@wmitsuda Rebased to develop and implemented one suggestion from Copilot (though with my own implementation).

Tested and all seems working for me 🙌 

Screenshot 2025-09-02 at 11 56 27 PM

src/execution/components/DecoratedAddressLink.tsx
```diff
@@ -145,9 +148,47 @@ const ResolvedAddress: FC<ResolvedAddressProps> = ({
...
+        style={
+          !dontOverrideColors
+            ? {
+                color: "var(--color-kleros-tag)",
+              }
+            : undefined
+        }
+        onMouseOver={
+          !dontOverrideColors
+            ? (e) =>
```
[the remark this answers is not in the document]
sealer3: I agree with this - please use [Tailwind class names](https://tailwindcss.com/docs/hover-focus-and-other-states) to change the color on hover.

wmitsuda:
@gmkung pls take a look at the previous review comments as well

wmitsuda:
I gave some thought about the tab vs. non tab question, let's make it another tab, here is my reasoning:

- Adding all those info from Kleros makes the Overview tab too busy and pushes the search results down by a lot IMHO, requiring some scrolling even on desktop.
- All those info packed into section in the second column looks like a "too busy" region in the page, I think we could workaround it by giving some extra padding, but in the end I think it is still too much nesting.

So, my proposal is:

- Move it to another tab between "Overview" and "Contract" called "Info" with the Kleros icon (like we have the Sourcify icon in the "Contract" tab.
- That tab should not be shown if there is no Kleros info for that address.
- Let's unfold all that info which is currently packed under the right panel of the "Verified by Kleros Scout" into their own fields.

image

wmitsuda:
I think we can then move the "Kleros Scout" link to the first field, i.e.:

Name (Kleros Scout):                  Maker DAO

wmitsuda:
I'd like to remove the blue tag which is next to the address, because it is drawing too much the attention and it is repeating what is already shown below ("MakerDAO: DAI stablecoin")

gmkung:
@wmitsuda Ok all changes done:
1. Removed the KlerosTag from AddressSubtitle
2. Moved Kleros Info to a separate tab
3. Expanded the info to occupy more vertical space for readability
4. Moved 'Verified by Kleros Scout' to the bottom of the tab.

wmitsuda:
Small things:

- The Kleros icon in the "Info" tab seems misaligned (compare it with the Sourcify icon/tab to see the vertical alignment)
- There is no horizontal lines dividing the sections (see the other tabs, let's use the same dividers/spacing pls)
- On verified domains list, change the space-y to 1 instead of 0.5.
- Can we change "Verify on Kleros Scout" to just "Kleros Scout" and move it together with the "Name" label? I was thinking about "Name (Kleros Scout):". The link alone at the bottom looks at bit odd.

image

image

wmitsuda:
let's also change the token formatting to be consistent with our standard formatting:

image

image

- let's use the format: "name (symbol) - N decimal(s)"
- the icon/text vertical alignment seems not baseline aligned (see other places where icon/text is used; text seems a bit up in the example)

gmkung:
Screenshot 2025-09-03 at 7 58 48 PM
Screenshot 2025-09-03 at 8 02 21 PM

Addressed all 6 points! For the token logo, it looks right to me now at the current height, but do you want it higher relative to the text? @wmitsuda

wmitsuda:
> Screenshot 2025-09-03 at 8 02 21 PM

there is an extra ":" in the "Name..." label

> Addressed all 6 points! For the token logo, it looks right to me now at the current height, but do you want it higher relative to the text? @wmitsuda

actually, I just compared it with the icon we are using in production and I think it is the same, so let's dismiss this one.

wmitsuda:
> @wmitsuda If there is Kleros information, I think it would be best to keep most of it on the address page, since it is important, and once ots2 is complete again, we'll already have enough ots2 tabs. That said, there's a lot we can do to compress the vertical space: as you mentioned, "View on Kleros Scout" can be turned into a vertical link or at least be placed on one of the previous lines, and token information can also be moved or removed.
> 
> In general the Copilot suggestions are accurate.
> 
> Also, before this is merged:
> 
> * [ ]  Remove AI-generated `kleros-test.md`
> * [ ]  Remove changes to `public/config.json`
> * [ ]  `supportedChains` config option is unused
> * [ ]  Move `kleros` config section underneath a new `dataSources` (or similar) root config key
> * [ ]  Move inline SVG files into external files with names

@gmkung pls don't forget to check our previous comments

gmkung:
Ok, removed the additional colon. @wmitsuda 
All previous changes and comments are already addressed I believe. Let me know if you spot any that's unfixed.

wmitsuda:
@gmkung it is not, see for example the change in public/config.json.

src/execution/components/DecoratedAddressLink.tsx
```diff
@@ -20,6 +22,16 @@ import { VerifiedContractRenderer } from "../address/renderer/VerifiedContractNa
...
 
+// Helper function to check if a Kleros tag has valid display data
+const hasValidKlerosData = (tag: KlerosAddressTag): boolean => {
+  return !!(
+    tag.project_name &&
+    tag.name_tag &&
+    tag.project_name.trim() !== "" &&
+    tag.name_tag.trim() !== ""
+  );
+};
```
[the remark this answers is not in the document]
wmitsuda: @gmkung let's solve this logic duplication

src/kleros/useKleros.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  if (!klerosConfig?.enabled || !address) {
+    return null;
+  }
+
+  const query = useQuery(
+    getKlerosAddressTagsQuery(
+      klerosConfig.enabled,
+      klerosConfig.apiUrl || "https://scout-api.kleros.link",
+      provider._network.chainId,
```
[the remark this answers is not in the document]
wmitsuda: @gmkung the suggestion from copilot is not good, but I think we should not be accessing the _network field here and in the next copilot comment.

gmkung:
@wmitsuda 
Actually these were addressed in commit 811fdc87d2a2de8b7ee756bb4eb35b0700850ddb and 3ade5c2651eded431e0e6e6303cd86fc3029bccb already:
>  Remove AI-generated kleros-test.md
>  Remove changes to public/config.json
>  supportedChains config option is unused
>  Move kleros config section underneath a new dataSources (or similar) root config key
>  Move inline SVG files into external files with names

The config.json file format is indeed not the same as originally, but is based on @sealer3 's advice in this [comment](https://github.com/otterscan/otterscan/pull/3407#issuecomment-3220319511). 
```json
{
  "erigonURL": "https://docs-demo.quiknode.pro",
  "assetsURLPrefix": "http://localhost:5175",
  "externalDataSources":
  {
      "kleros":
      {
          "enabled": true,
          "apiUrl": "https://scout-api.kleros.link"
      }
  }
}
```
Could you acknowledge the above? If you prefer me to move the externalDataSources config for Kleros somewhere else, let me know where you would prefer.

wmitsuda:
> @wmitsuda Actually these were addressed in commit [811fdc8](https://github.com/otterscan/otterscan/commit/811fdc87d2a2de8b7ee756bb4eb35b0700850ddb) and [3ade5c2](https://github.com/otterscan/otterscan/commit/3ade5c2651eded431e0e6e6303cd86fc3029bccb) already:
> 
> > Remove AI-generated kleros-test.md
> > Remove changes to public/config.json
> > supportedChains config option is unused
> > Move kleros config section underneath a new dataSources (or similar) root config key
> > Move inline SVG files into external files with names
> 
> The config.json file format is indeed not the same as originally, but is based on @sealer3 's advice in this [comment](https://github.com/otterscan/otterscan/pull/3407#issuecomment-3220319511).
> 
> ```json
> {
>   "erigonURL": "https://docs-demo.quiknode.pro",
>   "assetsURLPrefix": "http://localhost:5175",
>   "externalDataSources":
>   {
>       "kleros":
>       {
>           "enabled": true,
>           "apiUrl": "https://scout-api.kleros.link"
>       }
>   }
> }
> ```
> 
> Could you acknowledge the above? If you prefer me to move the externalDataSources config for Kleros somewhere else, let me know where you would prefer.

I think kleros is enabled by default with these flags, no? then the externalDataSources can be ommited.

Also, see the erigonURL, pls don't change it.

gmkung:
Ah yes the ErigonURL was indeed changed, I've reverted that now. 

Kleros is actually not enabled by default without config.json's externalDataSources right now. Do you want me to remove that config and just make Kleros enabled by default and hardcode the Kleros URL?

wmitsuda:
yes, let's make it on by default, but allowing the user to override the setting using the configuration

wmitsuda:
@gmkung pls read the comments above in this PR, I've pointed out a couple of changes they are all in this PR ☝️

also, make sure to format the code properly, see the autofix.ci job is fixing your code every time, you don't need to force-push every time once you fix it.

src/execution/components/DecoratedAddressLink.tsx
```diff
@@ -93,7 +106,7 @@ const DecoratedAddressLink: FC<DecoratedAddressLinkProps> = ({
           <FontAwesomeIcon icon={faCoins} size="1x" />
         </span>
       )}
-      {match && (
+      {match && (!klerosTags || klerosTags.length === 0) && (
```
sealer3: @wmitsuda mentioned before:

> Another thing I'd like to keep is having the Sourcify icon prepended to the control when the contract is verified by Sourcify, e.g., in the USDC contract; the contract is both verified by Sourcify and curated by Kleros.

So, this particular change should be removed.

public/config.json
```diff
@@ -1,4 +1,12 @@
 {
   "erigonURL": "http://localhost:8545",
-  "assetsURLPrefix": "http://localhost:5175"
+  "assetsURLPrefix": "http://localhost:5175",
+  "externalDataSources":
+  {
+      "kleros":
```
sealer3: These changes to the config should be removed. Instead, `useKlerosAddressTags` should incorporate the defaults: if the `"kleros"` key doesn't exist, for example, it should treat it as enabled with `"https://scout-api.kleros.link"` as the `apiUrl`.

src/components/InfoRow.tsx
```diff
@@ -2,11 +2,19 @@ import React from "react";
 
 type InfoRowProps = React.PropsWithChildren<{
   title: React.ReactNode;
+  noColon?: boolean;
```
sealer3: Remove the noColon changes to this file. I don't think you use `noColon` anyway.

src/kleros/useKleros.ts
```diff
@@ -0,0 +1,163 @@
...
+  const klerosConfig = config.externalDataSources?.kleros;
+
+  if (!klerosConfig?.enabled || addresses.length === 0) {
+    return null;
+  }
+
+  const query = useQuery(
+    getKlerosAddressTagsQuery(
+      klerosConfig.enabled,
+      klerosConfig.apiUrl || "https://scout-api.kleros.link",
```
sealer3: Please define the default Kleros API URL in one location (e.g., `export const klerosDefaultApiUrl = "https://scout-api.kleros.link";` and import it from other files as needed.

sealer3 (changes requested):

gmkung:
@wmitsuda @sealer3 I indeed missed a few things - apologies! 
Just took time to go through every comment and patched all the gaps that still remains (see commits above).
This is a summary of all the things fixed thus far:
```
🔧 Config & Structure
[x] Remove kleros-test.md
[x] Delete unused supportedChains
[x] Move kleros config under externalDataSources
[x] Kleros enabled by default (configurable via externalDataSources.kleros)
[x] public/config.json kept minimal (no Kleros block; defaults handled in code)
[x] Reverted unnecessary InfoRow changes (e.g. noColon). 
[x] Show Sourcify logo in DecoratedAddressLink if code is verified by Sourcify
🎨 UI/UX & Styling
[x] Icons: always show Sourcify icon when verified; keep Kleros icon/text styling
[x] Tabs: Kleros moved to dedicated “Info” tab with KlerosLogo
[x] External links: use ExternalLink for Kleros Scout + dapp URLs; “Kleros Scout” link placed in the “Name (Kleros Scout)” label
[x] Verified domains: unfurled list with proper spacing and ExternalLink on each domain
[x] Text/spacing: consistent ContentFrame tabs; no scrollbars; integrated label to reduce clutter
[x] Icon alignment: Kleros and Sourcify icons standardized to 16px and aligned
📝 Data Handling
[x] Optional fields handled gracefully (conditional rendering by field)
[x] “Public name tag” required: if missing, treat as no Kleros info (gated by hasValidKlerosData/hasValidKlerosTags)
[x] Checksum handling fixed (case-insensitive address lookup in Kleros response)
🗂️ Code & Assets
[x] Kleros SVGs externalized; fallback token icon externalized and always rendered to avoid layout shifts
[x] TypeScript/lint clean (no invalid ExternalLink props; repo-wide Prettier run)
[x] Prevent layout shifts: always-rendered token icon with fallback; no orphaned punctuation
♻️ Readability/DRY/KISS improvements
[x] Centralized helpers in src/kleros/useKleros.ts: hasValidKlerosData, hasValidKlerosTags, formatKlerosName, useKlerosConfig
[x] Removed duplicate Kleros fetch in DecoratedAddressLink
[x] Tightened types for Kleros config access
[x] Added concise JSDoc for exported helpers/hooks
```
I'm sure there must be still some things I missed, but I hope most of the outstanding things are resolved!

gmkung:
Just rebased to the latest origin/develop branch.

src/execution/address/AddressTransactionResults.tsx
```diff
@@ -13,6 +13,7 @@ import NativeTokenAmountAndFiat from "../../components/NativeTokenAmountAndFiat"
 import StandardScrollableTable from "../../components/StandardScrollableTable";
 import StandardTBody from "../../components/StandardTBody";
 import TransactionLink from "../../components/TransactionLink";
+
```
sealer3: Remove unnecessary change (imports are organized automatically)

src/kleros/useKleros.ts
```diff
@@ -0,0 +1,214 @@
...
+  if (addresses.length === 0) {
+    return null;
+  }
+
+  try {
+    // Use proxy in development, direct API in production
+    const isDevelopment = import.meta.env.DEV;
+    const endpoint = isDevelopment
+      ? "/api/kleros/api/address-tags"
+      : `${apiUrl}/api/address-tags`;
```
sealer3: We should avoid splitting logic between environments. The only other place we do something like this is when loading a config.

Is the purpose of this particular dev environment check only because Kleros' servers aren't returning appropriate headers (`Access-Control-Allow-Origin`) and we need to proxy the requests? **Kleros really should change its server response headers** if that's the case, since they are not configured to be friendly for API requests, and **the default Kleros server won't work in a production build** as-is.

> Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://scout-api.kleros.link/api/address-tags. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 204.

As a comparison, Sourcify's API server responses include the `Access-Control-Allow-Origin: *` header when you access them, like at <https://sourcify.dev/server/v2/contracts/1?sort=asc&limit=5>

sealer3 (changes requested):

gmkung:
I will test again after we make the change to the CORS headers for our API service. Please hold on.

gmkung:
@sealer3 ok we fixed the CORS issue on the server-side. Please have a look now!

vite.config.ts
```diff
@@ -13,4 +13,5 @@ export default defineConfig({
     imagetools(),
     tailwindcss(),
   ],
+  server: {},
```
sealer3: Undo this change please.

public/config.json
```diff
@@ -1,4 +1,4 @@
 {
-  "erigonURL": "http://localhost:8545",
+  "erigonURL": "https://docs-demo.quiknode.pro",
```
sealer3: Please make sure to undo this change. (You can set the config on your local machine for testing using a [`.env.development.local` file](https://docs.otterscan.io/config/reading/devel)).

sealer3 (changes requested):

sealer3:
> @sealer3 ok we fixed the CORS issue on the server-side. Please have a look now!

Thanks! I haven't tried this out locally yet, but I can see the Kleros server responses are including the right headers. Great.

gmkung:
Let me know if any further issues surface during testing @sealer3 !

src/index.css
```diff
@@ -8,6 +8,10 @@
   --color-link-blue-hover: #0468ab;
   --color-verified-contract: #2b50aa;
   --color-verified-contract-hover: #26007b;
+  --color-kleros-gray: #6b7280;
+  --color-kleros-gray-hover: #4b5563;
+  --color-kleros-tag: rgb(107 114 128);
+  --color-kleros-tag-hover: rgb(75 85 99);
```
sealer3: I wonder if we can find a different color for Kleros-tagged addresses, since the gray color suggests the address link is disabled in my opinion: 
image

src/kleros/useKleros.ts
```diff
@@ -0,0 +1,211 @@
...
+  chainId: bigint | undefined,
+  addresses: ChecksummedAddress[],
+): UseQueryOptions<KlerosResponse | null> => ({
+  queryKey: ["kleros", chainId?.toString(), ...addresses],
+  queryFn: () => {
+    if (!enabled || !chainId) return null;
+    return fetchKlerosAddressTags(apiUrl, chainId.toString(), addresses);
+  },
+  staleTime: 5 * 60 * 1000, // 5 minutes
+  gcTime: 10 * 60 * 1000, // 10 minutes
```
sealer3: I think we can increase these times by a lot. I noticed there was a refetch to Kleros after switching back to a tab that I hadn't visited after a number of minutes (< 15 minutes), and that's probably not necessary. How often do Kleros tags get updated?

src/index.css
sealer3: Also, looking at the above image, it looks like the address attributes (the `[C]`) is pushed below the baseline of the address label text in the Kleros label. That will have to be fixed.

gmkung:
@sealer3 
1. Proposed a new colour for the Kleros text 
2. Aligned the [C]'s vertical alignment
3. Updated the tags cache times to higher values. Technically tags can be added anytime, but the new values are more realistic and reduces redundant fetches.
Screenshot 2025-09-14 at 10 32 45 PM

sealer3:
Thank you. @wmitsuda could you review the color choice?

wmitsuda:
hi, sorry for the delay, I actually liked the previous color bc it was a black-ish that recalls to Kleros logo.

the new blue/purple-ish does not connect to the Kleros brand IMHO, it feels more a Sourcify thing.

can we find something more Kleros-black-ish that doesn't feel disabled?

wmitsuda:
also, I feel the new color stands out too much (both in light and dark modes). we should find something more balanced that doesn't make those labels look more important than others UI elements.

gmkung:
Screenshot 2025-09-30 at 5 59 21 PM
@wmitsuda @sealer3 how about now? Hopefully this is a good balance between all the comments so far.

sealer3:
Yes, this darker color looks better:

image

@wmitsuda With your approval, I think it's good to go now.

sealer3 (approved):

wmitsuda (approved):
alright, let's merge it

wmitsuda:
actually, after I merged it, I noticed apparently it broke ENS resolution, see pic below. @sealer3 could you pls investigate?

image

wmitsuda:
interestingly, our cypress tests caught it, but only in the run after it was merged to develop.

sealer3:
Okay, I'll look into this. Considering there are more hooks being rendered, there is probably a hook somewhere inside an execution branch (which breaks one of the rules of React).

PR #3508, opened 2025-08-31
Save local verifications to storage and show check

### LINKED ISSUE #3478: Local storage caching of local verification info
I think a next step incremental improvement to the local verification could be to cache the verification status of at contract locally:

- probably in browser local storage, so that info survives browser restarts
- I think simply storing a tuple (origin, chain, contract) should be enough to uniquely identify the verification and make it tamper-resistant
  - origin: the data source URL; usually the Sourcify API servers, but that should prevent collision/tampering/etc if the user mixes another Sourcify instance
  - chain: chain ID because contract addresses may vary between chains
  - contract: the address itself
- I think there should be ways to an attacker to inject a "verification ok" status into a user browser by manipulating the local storage and we should warn that somehow, but so far I think it may be a safe feature in general
- then we can reuse that info everywhere the contract address is displayed by using some variation of the Sourcify icon (green Sourcify? Sourcify with small check overlay?).

We can put this together under the experimental flag.

sealer3:
Closes #3478

In progress. Currently shows a checkmark next to addresses whose displayed metadata has passed a local verification:

image

Maybe the checkmark can replace the Sourcify logo.

### DIFF
--- a/src/components/Alert.tsx
+++ b/src/components/Alert.tsx
@@ -4,17 +4,19 @@ import React from "react";
 
 interface AlertProps {
   className?: string;
+  margin?: string;
   icon?: IconDefinition;
   children?: React.ReactNode;
 }
 
 const Alert: React.FC<AlertProps> = ({
   className = "bg-orange-100 border-orange-500 text-orange-700",
+  margin = "mb-4 mt-2",
   icon,
   children,
 }) => (
   <div
-    className={`${className} border-l-4 px-3 py-2.5 mb-4 rounded-sm mt-2 border-t-0 border-b-0`}
+    className={` ${className} ${margin} border-l-4 px-3 py-2.5 rounded-sm border-t-0 border-b-0`}
     role="alert"
   >
     {icon && <FontAwesomeIcon icon={icon} className="mr-2" />}
--- a/src/execution/AddressMainPage.tsx
+++ b/src/execution/AddressMainPage.tsx
@@ -12,6 +12,7 @@ import { useProxyAttributes } from "../ots2/usePrototypeTransferHooks";
 import SourcifyLogo from "../sourcify/SourcifyLogo";
 import { Match, useSourcifyMetadata } from "../sourcify/useSourcify";
 import { useWhatsabiMetadata } from "../sourcify/useWhatsabi";
+import { useIsLocallyVerified } from "../storage/CheckedContractStorage";
 import { ChecksummedAddress } from "../types";
 import { hasCodeQuery } from "../useErigonHooks";
 import { useAddressOrENS } from "../useResolvedAddresses";
@@ -23,13 +24,18 @@ const ProxyTabs: React.FC<AddressAwareComponentProps> = ({ address }) => {
   const { addressOrName } = useParams();
   const { provider } = useContext(RuntimeContext);
   const proxyAttrs = useProxyAttributes(provider, address);
+  const isLocallyVerified = useIsLocallyVerified(
+    proxyAttrs.proxyMatch,
+    provider._network.chainId,
+    address,
+  );
   return (
     <>
       {proxyAttrs.proxyHasCode && proxyAttrs.proxyMatch && (
         <NavTab href={`/address/${addressOrName}/proxyLogicContract`}>
           <span className={`flex items-baseline space-x-2`}>
             <span>Logic Contract</span>
-            <SourcifyLogo />
+            <SourcifyLogo locallyVerified={isLocallyVerified} />
           </span>
         </NavTab>
       )}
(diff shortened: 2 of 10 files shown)


wmitsuda:
I was actually thinking about embedding this info in the Sourcify logo itself, maybe adding a small overlay to the Sourcify logo, in order to save horizontal space.

wmitsuda:
I asked AI to simulate it, WDYT? I think we don't need to modify the Sourcify logo, but should be possible to use styling to create the effect.

image

sealer3:
Good idea, I'll do this.

sealer3:
Here's what it looks like now:

image

It looks like the metadata hash comparisons aren't always working for me, so the checkmark does not always show, or it shows and later stops being shown. I'm looking into that. (The issue is overwriting the sources key during local contract verification changes the hash and additionally changes the cached version.)

wmitsuda:
still a little big, could you shrink the checkmark to be ~1/3 of the sourcify logo dimensions? that's approximately the size in the AI-generated simulation above, and it looks a better.

also, the checkmark look black in dark mode

image

sealer3:
I shrank it to about 1/2 the Sourcify logo size, and in my opinion it's not very noticeable anymore. The dark mode checkmark should be fixed now.

image

sealer3:
The contract verification status is now also shown in the tab: 
image

Proxy contracts will theoretically also show that, but those can't be tested yet since it relies on ots2.

TODO:
- [x] Update tab when the current contract is verified
- [x] ~Do not display sources which don't match the metadata hash~ Display alert if there's a mismatch
  - [x] All contracts or just locally verified ones?

wmitsuda:
> The contract verification status is now also shown in the tab: image
> 

it looks like it broke the vertical alignment

wmitsuda:
using the browser built-in page zoom, it also looks like some of the white background "trick" is leaking some pixels

image

sealer3:
Yes, I suppose with different sizes they should have different absolute positions... I'll fix it. FontAwesome has support for [stacking](https://docs.fontawesome.com/web/style/stack), but not at the sizes we need, which is why we have to do it this way. I tried using stacking at equal sizes and got a similar effect around the border of the checkmark circle.

sealer3:
Fixed: 
image

And I also prevented the Sourcify logo from shrinking in the Contract tab when there wasn't space, which looked strange.

wmitsuda:
what about:

```
          <FontAwesomeIcon
            className="absolute bottom-[0.4px] right-[-1.2px] ml-1 text-white dark:text-black text-[4.6pt]"
            icon={faCircle}
          />
          <FontAwesomeIcon
            className="absolute bottom-0 right-[-1.5px] text-emerald-500 text-[5pt]"
            icon={faCheckCircle}
          />
```

in order to cover a little less the Sourcify logo?

wmitsuda:
image

sealer3:
I'm of the opinion that especially for non-high-res or small displays (remember when we began optimizing for mobile devices), that is going to be too small to notice. I have to look close to the icon to notice the checkmark at that size. Among these I still prefer the check-circle that covers up the Sourcify logo the most: https://github.com/otterscan/otterscan/pull/3508#issuecomment-3242846470

sealer3:
For reference, look at how much Sourcify's own icon covers up the logo:
image

https://repo.sourcify.dev/favicon-verified.ico

sealer3:
Rather than not showing a source at all if its hash doesn't match the one in the metadata, I added an alert:

image

wmitsuda:
> For reference, look at how much Sourcify's own icon covers up the logo: image
> 
> https://repo.sourcify.dev/favicon-verified.ico

notice there is a reason for that specific case: it is the favicon and it is constrained to fit everything in a square.

in our case we can let the overlay overflow slightly to the right (both the original and my suggested small version does that) making the combination actually a rectangle.

should we then maybe use the "squared" version from Sourcify (we can ask for a highres bc that one is an .ico) in order to preserve the aspect ratio?

sealer3:
How about this one? I matched the size ratio between the checkmark circle and the Sourcify logo to the one present in the Sourcify icon. I then moved the checkmark circle further right.

image

In making the logo non-square, at least one of the following will happen: (1) we make the width different from the non-checkmark logo, and contract titles won't line up when one is locally verified and the other isn't - this might be fine, or (2) we keep the width the same as with the non-checkmark logo and the right margins will be different.

After looking at your version again, I like it more than the above one, but in both of these it's hard to notice:
image

The checkmark dominates in the original version, which I still prefer:
image

If you'd like to ask Sourcify for the verified icon, you can, but when we do it ourselves then we get consistency with the checkmarks shown next to the matches after local verification.

Honestly I could just go with your suggested version if you like. It's reasonable, if small.

wmitsuda:
let's go with mine then, we can always revisit those decisions later

sealer3:
All icons are now updated immediately when a contract is locally verified (using React Query as the state invalidation service). Ready for review.

PR #3593, opened 2025-09-30
Add custom output decoding button and field

### LINKED ISSUE #3100: Add manual parsing type dropdown
A user on Discord suggested we add a dropdown for parsing function return data when the type is not known:

Image

Image

If `Raw` is the only option for decoding, we can include a dropdown as suggested. It also might be feasible to add an input box if the user wants to enter a type manually in order to parse complex types.

sealer3:
Closes #3100. Adds a Custom tab with an input field for the type, with two autocomplete-friendly entries (`address` and `uint256`). Currently, this only supports decoding outputs, but in the future we could support inputs too. Multiple outputs can be specified via a tuple.

Examples on mainnet:

`/address/0x4025ee6512DBbda97049Bcf5AA5D38C54aF6bE8a/readContract`
image

`/tx/0x01cebe392b95513f76bd9e94caba96d0b623f2538d483d1c790f0964f00e6920/trace` (multiple outputs supported via tuple type)
image

### DIFF
--- a/src/execution/transaction/decoder/OutputDecoder.tsx
+++ b/src/execution/transaction/decoder/OutputDecoder.tsx
@@ -64,8 +64,11 @@ const OutputDecoder: React.FC<OutputDecoderProps> = ({
     setCustomParamTypes(parsed);
 
     try {
-      const decoded = AbiCoder.defaultAbiCoder().decode(parsed, data);
-      setCustomDecoded(decoded);
+      const decodedProxy = AbiCoder.defaultAbiCoder().decode(parsed, data);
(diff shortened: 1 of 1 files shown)


wmitsuda (approved):
I'll merge it, but just a comment: the first example doesn't bring me anything bc it can't find the contract on Sourcify.

so you either pasted the wrong address or there was some regression

sealer3:
@wmitsuda The point is that the contract doesn't exist on Sourcify, so this is the Read Contract page using the whatsabi-found function. You must have 4bytes enabled and this function included in your 4bytes folder for it to show.

PR #3633, opened 2025-10-19
Limit formatValue decimals to 80

### LINKED ISSUE #3631: Tokens with more than 80 decimal places throw ethers FixedNumber error
It seems like this issue can be mitigated by checking the decimals in the `formatValue` function in `formatter.ts`.

sealer3:
Fixes #3631, tested on the token holdings page which caused that issue.

Note: 10^80 > 2^256.

### DIFF
--- a/src/components/formatter.ts
+++ b/src/components/formatter.ts
@@ -2,7 +2,7 @@ import { BigNumberish, formatUnits } from "ethers";
 import { commify } from "../utils/utils";
 
 export const formatValue = (value: BigNumberish, decimals: number): string => {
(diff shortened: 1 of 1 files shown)


wmitsuda:
so, we assume those are buggy tokens?

sealer3:
Yes, ethers does not currently support formatting values with more than 80 decimal places. Since you would not even be able to represent `1` in a uint256 for a token with more than 80 decimals, it's unlikely that a normal token contract would ever do that. We'll just show the value without decimals in that case (instead of, for example, a malicious or meme token taking down the whole UI).

wmitsuda (approved):

PR #3798, opened 2026-04-27
Use the ENS Universal Resolver for name resolution

Dhaiwat10:
Hey! I'm Dhaiwat from the DevRel team at ENS Labs. We're working on getting libraries across the ecosystem ready for ENSv2.

## What

Routes ENS resolution through the Universal Resolver (`0xeeeeeeee14d718c2b47d9923deab1335e144eeee`) instead of the legacy ENS Registry.

ENS resolution in this repo goes through `ethers` v6 at:
- `src/useResolvedAddresses.ts` — `provider.resolveName()`
- `src/api/address-resolver/ENSAddressResolver.ts` — `provider.lookupAddress()`

ethers v6 is not ENSv2-aware natively, so this PR installs `@ensdomains/ethers-patch-v6` and imports it as the first import in `src/index.tsx` so the patch loads before any ethers usage.

## Why

The legacy registry path returns stale results for names managed by the new Universal Resolver. The UR is the canonical entry point for ENS resolution on Mainnet — viem, ensjs, and patched ethers all use it.

More: https://docs.ens.domains/web/ensv2-readiness

## Verification

- `ur.integration-tests.eth` should resolve to `0x2222222222222222222222222222222222222222` (legacy returns `0x1111111111111111111111111111111111111111`).

### DIFF
--- a/cypress/e2e/mainnet/ens-universal-resolver.cy.ts
+++ b/cypress/e2e/mainnet/ens-universal-resolver.cy.ts
@@ -1,0 +1,15 @@
+describe("ENS Universal Resolver", () => {
+  it("Should resolve ur.integration-tests.eth", () => {
+    // ENS publishes ur.integration-tests.eth with a wildcard CCIP-Read resolver
+    // that legacy resolution can't reach — only the Universal Resolver returns
+    // an address. If this passes, the patch is wired up.
+    cy.visit("/");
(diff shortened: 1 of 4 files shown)


Dhaiwat10:
Marking this as a draft until CI is run and it passes :)

PR #3802, opened 2026-08-21
Batch Kleros address tag lookups into a single API request

### LINKED ISSUE #3801: Batch Kleros address tag lookups instead of per-address requests
## Problem

`useKlerosAddressTags` in `src/kleros/useKleros.ts` fetches address tags one address at a time. Each `DecoratedAddressLink` rendered on screen triggers an independent `POST /api/address-tags` request with a single-element array (`addresses: [address]`).

On a block page with 25 transactions, this results in up to 50 parallel HTTP requests (one per unique from/to address). Transaction detail pages with many internal operations can trigger even more.

## Current state

The underlying fetch function `fetchKlerosAddressTags` already accepts an array of addresses. A batch hook `useKlerosAddressTagsBatch` is already implemented (line ~178) but is never imported or used by any component.

## Expected behavior

Address tag lookups should be batched — collect all visible addresses on the page and fetch their tags in a single (or few) API call(s) instead of N individual requests.

## Suggested approach

1. Use `useKlerosAddressTagsBatch` (or a similar batching strategy) at the page/list level to prefetch tags for all addresses in the current view.
2. Distribute results to individual `DecoratedAddressLink` components via context or cache population.
3. React Query's `staleTime` (currently 1 hour) already handles deduplication of repeated addresses — the batch just needs to seed the cache.

## Relevant code

- `src/kleros/useKleros.ts` — `useKlerosAddressTags` (per-address, used everywhere) vs `useKlerosAddressTagsBatch` (batch, unused)
- `src/execution/components/DecoratedAddressLink.tsx` — triggers per-address fetch on every address link render

salgozino:
Each DecoratedAddressLink triggered an independent POST /api/address-tags request, producing up to 50 parallel requests on a block page (2 per transaction). The upstream API already accepts arrays of addresses.

This adds a transparent microtask batcher inside useKlerosAddressTags:
- Addresses registered during a render frame are coalesced and fetched in one batch request
- Individual React Query cache entries are seeded from the batch response so each consumer hook reads its data with zero component changes
- Fresh cache entries are skipped; failed batches are negative-cached for 30s to avoid retry loops
- Removes the unused useKlerosAddressTagsBatch hook (dead code with a Rules of Hooks violation) now that batching is transparent

Fixes #3801

### DIFF
--- a/src/kleros/useKleros.ts
+++ b/src/kleros/useKleros.ts
@@ -1,7 +1,14 @@
 import { useQuery, type UseQueryOptions } from "@tanstack/react-query";
-import { useContext } from "react";
+import { useContext, useEffect } from "react";
+import { queryClient } from "../queryClient";
 import { ChecksummedAddress } from "../types";
 import { RuntimeContext } from "../useRuntime";
+
+/** Cache duration for a successful Kleros response (1 hour). */
+export const KLEROS_STALE_TIME = 60 * 60 * 1000;
+
+/** Cache duration for a failed/empty Kleros response (30 seconds). */
+export const KLEROS_ERROR_STALE_TIME = 30 * 1000;
 
 /** True if a single Kleros tag has both non-empty project_name and name_tag. */
 export const hasValidKlerosData = (tag: KlerosAddressTag): boolean => {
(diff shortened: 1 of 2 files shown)


src/kleros/useKleros.ts
```diff
@@ -174,39 +296,3 @@ export const useKlerosAddressTags = (
 
   return addressKey ? addressResponse[addressKey] : null;
 };
-
-/** Fetch Kleros tags for multiple addresses; useful for lists. */
-export const useKlerosAddressTagsBatch = (
```
salgozino, line 179: Removed because was not used and now we are batching inside of `useKlerosAddressTags`

